PR #32, opened 2018-09-02
Fix build on linux

DamsRoxxx:
Since my fix of Windows build, libretro-puae could not build on Linux...

On linux we need to include <sys/timeb.h> when FTIME is not defined.


DamsRoxxx:
Travis linux build don't work...

Strange on my linux box it work...

Will try another fix.

PR #61, opened 2019-04-20
Add IPF support through capsimg

escalade:
This compiles and links in the internal capsimg code. By default this tries to dlopen libcapsimage.so.2. Tested working with Frode Solheim's capsimg fork. I don't think there's any license issue as the capsimage code itself is external.

PR #62, opened 2019-05-30
IPF support through CAPSIMG

### LINKED ISSUE #43: Improvement of IPF Support
IPF are original copies from amiga games. However it seems there are several issues when loading an ipf with puae. Either there is not enough memory or there is a software issue. Normally IPFs should behave like the ADF file.

DamsRoxxx:
@escalade activated support in #61 but the code was only for linux so windows build was broken.

Ported some functions from https://github.com/FrodeSolheim/fs-uae to make this work on windows too.

Added documentation of IPF support in README.md

Fix #61 
Fix #43

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -26,6 +26,7 @@ SOURCES_C +=  $(EMU)/main.c \
 				  $(EMU)/keymap/keymap.c\
 				  $(EMU)/diskutil.c \
 				  $(EMU)/caps/caps.c \
+				  $(EMU)/caps/uae_dlopen.c \
 				  $(EMU)/zfile.c \
 				  $(EMU)/zfile_archive.c \
 				  $(EMU)/cfgfile.c \
--- a/README.md
+++ b/README.md
@@ -74,6 +74,15 @@ It is critical to use kickstarts with the right MD5, otherwise the core might no
 |kick40063.A600|Kickstart v3.1 (Rev. 40.063)|Amiga 600|e40a5dfb3d017ba8779faba30cbd1c8e|
 |kick40068.A1200|Kickstart v3.1 (Rev. 40.068)|Amiga 1200|646773759326fbac3b2311fd8c8793ee|
 
+### IPF Support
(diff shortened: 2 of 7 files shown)


5schatten:
The core tries to open `libcapsimage.so.2` while packages like [capsimg](https://github.com/FrodeSolheim/capsimg/blob/master/Makefile.fs#L49) provide `libcapsimage.so.5.1`
https://github.com/libretro/libretro-uae/blob/master/sources/src/caps/caps.c#L30 so wouldn't it make sense to use `libcapsimage.so` instead?

DamsRoxxx:
> The core tries to open `libcapsimage.so.2` https://github.com/libretro/libretro-uae/blob/master/sources/src/caps/caps.c#L30 while packages like [capsimg](https://github.com/FrodeSolheim/capsimg/blob/master/Makefile.fs#L49) provide `libcapsimage.so.5.1`
> so wouldn't it make sense to use `libcapsimage.so` instead?

In fact, capsimg provides capsimg.so. In your link look at this [line](https://github.com/FrodeSolheim/capsimg/blob/master/Makefile.fs#L49)
```
cp CAPSImg/libcapsimage.so.5.1 capsimg.so
```

So, in this PR, I use `capsimg.so` (and  `capsimg.dll` for windows) provided by capsimg instead of `libcapsimage.so.2`. 

See https://github.com/DamsRoxxx/libretro-uae/blob/8625835aea844181cbb3c94e0c4c02e980353561/sources/src/caps/caps.c#L33

5schatten:
It only provides capsimg.so if you use the recent created Makefile.fs which does not really work that well if you cross compile. If you use bootstrap.fs & configure.fs then you'll get a Makefile like this https://pastebin.com/dVNvGedf but basically the point is that `libcapsimage.so.2` is actually only valid for an outdated capsimg version. So IMHO it should open `libcapsimage.so` which is in line with the common lib name scheme.

DamsRoxxx:
I just used the name `capsimg.so` because I found them in the package provided by fs-uae at https://fs-uae.net/download#plugins

I think lambda users will download that package (I've provided the link in the updated Readme.md). On Windows it work seemlessly. Don't know for linux.

But, sure, It can be changed for whatever possible. Don't know what's the better...

DamsRoxxx:
Hi @twinaphex, could you please merge ?

DamsRoxxx:
Hi @hizzlekizzle & @twinaphex  could you please merge ?

Issue #28, opened 2018-05-07
2-Player Support


qclart:
2-Player is not supported.  

I've added a 2-Player hack to my fork - https://github.com/qclart/libretro-uae 

I will do a pull request if I find a more elegant solution or if this hack proves stable.

escalade:
Any progress on this?

PR #65, opened 2019-05-31
Add support for a second controller and remap keys

escalade:
It looks like the author went inactive, so I took the liberty to merge in his second controller commits:

https://github.com/libretro/libretro-uae/issues/28
https://github.com/qclart/libretro-uae

It was done quickly by hand and I'm not sure if I got it correctly. It does work great for me on the few games I tested though.

Probably best if someone who actually knows what they're doing reviews the code.

The new keymap is as follows:

L2 = show/hide status
R2 = UNUSED
L = mouse speed down (min 1)
R = mouse speed up (max 6)
SEL = toggle mouse/joy mode
STR = show/hide virtual keyboard (vkbd)
A = joystick fire 1/mouse 1/vkbd key pressed
B = joystick fire 2/mouse 2
X = UNUSED
Y = UNUSED

The "enter GUI" function was removed.

@qclart
@twinaphex

### DIFF
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -36,9 +36,9 @@ extern int CROP_HEIGHT;
 #define XSIDE  (CROP_WIDTH/NPLGN -1)
 #define YSIDE  (CROP_HEIGHT/8 -1)
 
-#define YBASE0 (CROP_HEIGHT - NLIGN*YSIDE -8)
-#define XBASE0 0+4+2
-#define XBASE3 0
+#define YBASE0 (CROP_HEIGHT - NLIGN*YSIDE -80)
+#define XBASE0 12
+#define XBASE3 4
 #define YBASE3 YBASE0 -4
 
 #define STAT_DECX 120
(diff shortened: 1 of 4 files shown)


escalade:
Unfortunately, this segfaults on x86_64. ARM works good though.

DamsRoxxx:
> Unfortunately, this segfaults on x86_64. ARM works good though.

So it should not be merged ?

escalade:
Only if you want to break x86_64 :)

Too bad, hopefully someone knowledgeable will have a look and fix it as it actually makes this core useable.

escalade:
@rsn8887

Sorry for tagging you in this, but I noticed your superb work on the Vice core and a revamped virtual keyboard is just what this core needs too ;) The current one is absolutely hideous. Proper 2-player would be great as well! Feel free to ignore this, please don't misunderstand this as any kind of demand, thanks so much for the C64 love. Just seems like it's within your area of expertise and also interest.

sonninnos:
Cheers! I still haven't checked out what is the state of this UAE core, but I think I will after the Vice is core is so called "done".

escalade:
@sonninnos 

Awesome, great to see some love for the Commodore systems. I think this core runs quite well, just needs some polish and love like you have done for Vice. There's a rudimentary FS-UAE core as well, which I suppose could be brought up to speed with Frode's latest work and polished. Looks like you're an ARM guy though and although it runs I think PUAE is a better performer and runs A500/A1200 games like a charm. Anyway, thanks for the response and optimistic outlook :)

PR #74, opened 2019-08-17
Fix mingw cross compile and add some missing nintendo targets

webgeek1234:
Nintendo targets have only been tested as far as compile, they have not been tested on target hardware.


rsn8887:
Thanks!

PR #76, opened 2019-08-22
Core options for sound

sonninnos:
Removed unnecessary options and added a necessary option.

Sound options will come to effect immediately, and not just change the config for the future.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -286,6 +286,10 @@ static void update_variables(void)
 
    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
    {
+		strcat(uae_config, "sound_stereo_separation=");
+		strcat(uae_config, var.value);
+		strcat(uae_config, "\n");
+
 		int val = atoi(var.value) / 10;
 		changed_prefs.sound_stereo_separation=val;
    }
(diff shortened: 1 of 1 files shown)


rsn8887:
Awesome!

escalade:
Sound is all messed up in Battlesquadron intro (slow and missing some instruments?), so something went wrong here. I'm really excited that you guys are polishing up this core though :)

sonninnos:
Odd, I just compared with WinUAE and I didn't spot any slowness or missing instruments.

escalade:
There's definitely an issue, the offending commit is 8bc656ae887e77b02b879d01979b0cf160e93cf2

Launching IPF: MD5 7ba02ae220a84105a3bf788afd7b9f8e  Battle Squadron - The Destruction of the Barrax Empire! (Europe).ipf

rsn8887:
I suspect the changed sound filter defaults that were added?

rsn8887:
Probably the two filter options should be core options with previous defaults.

sonninnos:
Maybe perhaps, but I can't confirm nor fix since I can't reproduce.

Also try changing the default stereo separation to something smaller?

escalade:
I'll test some more later, it seems to be an issue in any game (confirmed with Dogs of War intro). Sounds a bit upspeed and some missing instruments, very weird. I will upload samples as well.

I suppose it could be specific to my environment somehow. I removed every default from the core options on launch before testing. I have 48000 as output rate and use Pulseaudio as backend.

rsn8887:
I will add the sound options back in so you can test.

sonninnos:
You should be able to narrow the culprit down with just copying the autocreated .uae from retroarch/saves/, removing the rows and launching the new file. They won't get overwritten by the core until you go and change core options.

rsn8887:
@escalade : I just added options for everything regarding sound instead of forcing any defaults. Should be in next nightly. Can you try if any sound options work for you? E.g. try "mixed" or "stereo"  , try "filter: enabled" "filter: off" etc.

The previous defaults that should to work for you for sure were:
`channels: stereo`
`filter: off`
`filter_type: standard`
`stereo_separation: i don't know`

escalade:
@rsn8887 
@sonninnos 

It was the stereo separation indeed. I found this post by Frode Solheim: http://eab.abime.net/showpost.php?s=81e9ebd09e9aa3d9ccf36ef110123aa3&p=862802&postcount=11

I've tried with 70% and it sounds good to me thanks. You're on a roll guys, keep the commits coming it's Amiga time 💃 Let me know if I can send beer money somewhere :)

PR #77, opened 2019-08-23
Major controls overhaul 

### LINKED ISSUE #28: 2-Player Support

### LINKED ISSUE #40: Mouse and Keyboard cannot be mapped to Controller Buttons
It i not possible to map Keyboard Keys or Mouse buttons to the Retropad.
One example would be to map Mouse Movements to the right analog stick.
Another one is to map the Enter on the keyboard to a controller button.

### LINKED ISSUE #42: Mouse & Keyboard Option for Controls
Adding an option to use mouse and keyboard instead of the retropad would be useful.

### LINKED ISSUE #51: Keyboard is set to AZERTY (Fr) by default. Additional Shift key problem.
Hello,

The keyboard has an AZERTY layout by by default without an "input.prefs" file and even when one with an English keymap is present. This happens with kickstart 3.1 and 1.3 alike, with or without startup-sequence. 
azerty

Image is from wikipedia:
500px-kb_france svg

The only temporary workaround is to counterintuitively choose an actual French layout in input settings under workbench. This way the layout switches to English. However it reverts back to French outside that workbench environment ie. when booting without startup sequence. Also this workaround doesn't fix the special characters.

fre
ame

Windows 10 country preferences were set to Canada when I first encountered this error but changing it to United States and restarting Windows did not resolve the problem. Windows display language was always set to "English (United States)". I haven't encountered this problem with any other software including WinUAE using the same hdf files and kickstart roms. Also, Canada does not use an AZERTY layout even for French.

region

**Shift problem:**
In addition to this, the shift key works as a toggle and not an actual modifier key. It gets stuck once pressed and has to be pressed again to switch back to normal typing similar to a caps lock but different because it also modifies the special characters that way, unlike a caps lock.

Thanks in advance and thank you for developing this core.

sonninnos:
And some bonus maintenance with libretro.h and core options.

Closes #42 
Closes #40 
Closes #28 
Closes #51


rsn8887:
Absolutely great!

escalade:
This made my day :)

PR #82, opened 2019-08-26
Floppysounds

escalade:
The default floppy sounds doesn't sound very good, I compared with the fs-uae samples and there's two higher quality samples.


sonninnos:
I feel the new ones are better too, so I vote yes.

escalade:
@sonninnos 

Hoping for a refreshed keyboard as well, the one from vice is really nice ;)

sonninnos:
> 
> 
> @sonninnos
> 
> Hoping for a refreshed keyboard as well, the one from vice is really nice ;)

Heh, patience, it is at the top of the list and I will get to it.

PR #91, opened 2019-08-30
Button layout according to Retro Standards

sonninnos:
And a transparency toggle of some kind for the vkbd as a bonus.

Also removed the joy/mouse-toggle from GUI and made it automatic.

### DIFF
--- a/README
+++ b/README
@@ -27,29 +27,14 @@ Credits to Mustafa 'GnoStiC' TUFAN
 
 And of course for the RetroArch/Libretro team : "http://www.libretro.org/"
 
-It's a quick hack.
-At this time , works the basis but buggy on win/linux/android.
-
-Remember ,Everything not working well, It's a debug release , so expect to more bug.
 
 Default Control :
 
-L2  Show/Hide statut
-R2  Sound on/off
-L   Toggle Num Joy .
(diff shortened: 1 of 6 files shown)


rsn8887:
Awesome, thanks!

MajorWilly:
@sonninnos I haven't updated yet, still waiting for the buildbot. I'm curious to know about this automatic mouse. What happens when you have a game that accepts mouse or joystick to control, like Indianapolis 500 ? I'm using a handheld that has joypad controls so will it default to mouse controls (which needs changing sensitivity) ?

Also, did you know that (with default controls) the Right Analogue Stick controls the mouse !?

sonninnos:
That change only applies to the GUI (which has very little or no use anymore), and the regular mouse already does this, as in if you move an actual physical mouse, it always works. So it changes nothing on the RetroPad mapping.

And sure I know about the right stick, I tweaked its sensitivity a bit lately. Only thing missing from that is to be able to press the mouse buttons without changing the retromode.

MajorWilly:
I understand now. I never realised there was a GUI ! It looks the same as Hatari which is the only part I saw on that core as it doesn't work on Android. No great loss as we now have this far superior Amiga core ! Thanks.

PR #94, opened 2019-08-30
Minor changes

sonninnos:
Now I promise this is the last one for today, heh.

Reason for the floppy fix was that I launched Push Over and noticed it always clicks the extra drive during gameplay for absolutely no reason, as it does not even use it for the second disk. Sure I could disable the sound emulation completely, but..


rsn8887:
Awesome! Merging...

PR #97, opened 2019-08-31
Core option update

sonninnos:
One missing compatibility option added and the rest got some lipstick.

Oops, missed a tiny spot, hence the force.


rsn8887:
Thanks!

PR #104, opened 2019-09-03
Fixes and finetunings

### LINKED ISSUE #85: Compatibility Issues with UAE config file
Tested following games and they do not boot:
North & South
The great Giana Sisters

sonninnos:
VKBD had a small oopsie.

Prevented special negative keys from being selectable in hotkeys.

Tuned uae config launch logic:
- Instead of launching the uae directly, it is appended to the usual temporary config, to effectively have the core options as a base, and then overridden with the custom configs.
- Disk and drive images use the same row, so they won't get replaced

Amirite?

Closes #85


rsn8887:
Thanks

PR #106, opened 2019-09-03
Renovations and fixes

### LINKED ISSUE #105: Bug if show_leds is set to false
If show_leds parameter is set to false or not specified (default false) some games hang at a moment or another.

For example Rick Dangerous 2 (adf or whd version) hang when selecting level.

sonninnos:
Closes #105


rsn8887:
Thanks!

PR #108, opened 2019-09-04
Removed GUI + remnants and added chipset_compatible

sonninnos:


rsn8887:
Looks good to me actually. Merging

PR #112, opened 2019-09-08
Forced bogomem_size to 0 by default, added A1200 model without extra memory, updated README accordingly

sonninnos:
This extra model allows Second Samurai AGA to work.

Also changed the pecking order of the filename detection to check A1200 models first, then A600 and then A500 models.

And since the UAE built in model prefs are never used with this core, I defaulted bogomem to 0. Otherwise A1200 would have bogomem, and that is no bueno.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,12 @@ The following models are provided (hardcoded configuration):
 
 |Model|Description|
 |---|---|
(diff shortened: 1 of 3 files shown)


rsn8887:
Awesome, thanks!

Issue #56, opened 2019-02-27
Request: resolution switching for crt switchres


retrorepair:
Now that retroarch is capable of switching resolution on the fly for crt users, it would be nice if we weren't limited to one resolution on boot when many games and apps switch res mid game.

This would mean either deciding which options are enabled (lores, linemode, pal, ntsc) on the fly based on the current framebuffer or using the framebuffer itself to fill the variables in retro system av info (forget the exact name).

Obviously this should be an option either in the core settings or config file as it won't apply to everyone.

retrorepair:
The benefit is CRT users will then have progressive and interlaced modes and any resolution that the virtual amiga draws will be displayed correctly without scaling issues.

retrorepair:
I should add this option is aimed mainly at CRT TV users so the native resolution is low res

retrorepair:
https://forums.libretro.com/t/15-khz-crt-documentation-wiki/15671/4

This may explain more as I've not worked on this project for over 6 months and it's early in the morning :)

rsn8887:
This should've just been fixed by the merging of PR #113 . If you think there's still some problem with this, please comment here, and I will re-open the issue.

PR #113, opened 2019-09-09
Dynamic PAL/NTSC switching

sonninnos:
Attempt at doing resolution and timing switching on the fly. It can be tested for example at Early Startup (hold LMB+RMB while booting A1200). Protracker also has a toggler.

If there is a way to achieve the same result without the desktop flashing, I'm all ears. I could not get the new timing to apply without using RETRO_ENVIRONMENT_SET_SYSTEM_AV_INFO, and that causes the flash. Geometry alone updates nicely without the flash.

If nothing else helps then maybe a core option should be added to disable the Hz change but allow the geometry change.

This will fix the double statusbar issue with Dyna Blaster (if the statusbar is at the bottom), but because it uses a weird 60Hz PAL mode (but acts like normal NTSC on a real Amiga), and it keeps returning to PAL for a split second during loadings for zero reason, it needed some trickery.

Also found more unused GUI stuff for removal.

### DIFF
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -18,7 +18,7 @@
 extern cothread_t mainThread;
 extern cothread_t emuThread;
 
-void retro_update_geometry(bool);
+bool retro_update_av_info(bool, bool, bool);
 
 #define LOGI printf
 
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -1,5 +1,5 @@
-#ifndef LIBRETRO_HATARI_H
-#define LIBRETRO_HATARI_H
+#ifndef LIBRETRO_GLUE_H
+#define LIBRETRO_GLUE_H
 
 #include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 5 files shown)


rsn8887:
I think instead of calling a function directly in custom.c, it might be better to just set a global variable currently_ntsc there, and then use that variable in retro_get_system_av_info() to set everything in that function. Then there will be no extra calls with RETRO_ENVIRONMENT_SET_XXX, only the necessary ones. I think retro_get_system_av is called automatically by the frontend, and it is called often enough for dynamic switching. At least that’s how it was in libretro Vice if I remember it correctly.

sonninnos:
Thanks, but unfortunately VICE does not do the Hz change correctly without restarting either. I use maximum skew to resample everything to the display Hz, and while changing PAL to NTSC does change the geometry, the drive sound pitch stays the same, when it should be repitched because of the skew. After restart the pitch is changed.

Looks like retro_get_system_av_info() will not get called after startup with RETRO_ENVIRONMENT_SET_GEOMETRY.

rsn8887:
Yeah I was wrong, you are doing this correctly...

sonninnos:
Finally ready for vigilant testing! It was a pickle if I may say so.

Looks like this closes issue #56 too?

rsn8887:
Merging, thanks!

PR #119, opened 2019-09-14
More core options to apply in realtime + crop overscan fix

### LINKED ISSUE #117: Crop overscan + statusbar at bottom = no statusbar
If you have 'Crop Overscan' ON and the statusbar at the bottom then you can no longer see the statusbar. It's OK at the top though.

NTSC screen is also cut off at the bottom. (Tested with Dynablaster). Is it stretching the screen and not centering ?

sonninnos:
Tried to make as many core options to work realtime as possible.
Also changed core option placements and fixed indentations.

Closes #117

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -75,6 +75,8 @@ extern void retro_joy(unsigned int, unsigned long);
 extern unsigned int uae_devices[4];
 extern int mapper_keys[30];
 extern int video_config;
+extern int video_config_aspect;
+extern bool request_update_av_info;
 extern bool opt_enhanced_statusbar;
 extern int opt_statusbar_position;
 extern unsigned int opt_analogmouse;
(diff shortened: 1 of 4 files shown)


sonninnos:
For some reason the vertical centering does not come into effect at the Dyna Blaster loading screen, but when the loading is finished. Good enough, I say. The centering does work fine in Early Startup, so I'll fix it some other time, maybe.

But since core options are supposed to work only in one direction, I didn't try to force centerings while crop overscan is on. But did include small prints to indicate that they are important to enable.

MajorWilly:
Nice. I'll say thanks now but will probably be a couple of days before I can try it having to wait for it to be pushed and then another day for the buildbot. I assume this fixes the stretched aspect ratio of NTSC or did that only happen with Dynablaster ?

sonninnos:
I did not see anything wrong with the aspect ratio with crop on or off.

Without crop:
retroarch_2019_09_15_01_47_49_747

With crop launched with (NTSC) in the filename:
retroarch_2019_09_15_01_47_29_959

sonninnos:
My CRT is busted, but this is how it looks on the real thing:
IMG_20190915_022325

sonninnos:
But indeed it does look very cropped with crop on, so I'll say just don't use crop with this game. Aspect ratio forcing won't help, since the non-NTSC area is not usable.

retroarch_2019_09_15_02_34_25_869

sonninnos:
I couldn't help myself. Since there already was an exception to detect the weird PAL60hz mode in Dyna Blaster I also added an exception for video size, cropped or not:
retroarch_2019_09_15_04_03_36_545

rsn8887:
Thanks, merging!

PR #123, opened 2019-09-17
Savestate support

### LINKED ISSUE #109: Save states ?
Just wondered if you have any plans to add save state support (for ADF/IPF) ?

sonninnos:
It uses a temp file like the VICE core, but the state restoring needs to be requested to be fired at the right time in m68k_go().

Only drawback for now seems to be rewinding, which is unusably slow. Still better than nothing.

Closes #109

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -11,6 +11,7 @@
 #include "uae.h"
 #include "options.h"
 #include "inputdevice.h"
+#include "savestate.h"
 
 #define EMULATOR_DEF_WIDTH 720
 #define EMULATOR_DEF_HEIGHT 568
@@ -92,6 +93,12 @@ extern int cd32_pad_enabled[NORMAL_JPORTS];
 
 int mapper_keys[31]={0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0};
 static char buf[64][4096] = { 0 };
+
+#ifdef WIN32
+#define DIR_SEP_STR "\\"
+#else
+#define DIR_SEP_STR "/"
+#endif
 
 static retro_video_refresh_t video_cb;
 static retro_audio_sample_t audio_cb;
@@ -1932,6 +1939,10 @@ void retro_init(void)
 	dc = dc_create();
 	environ_cb(RETRO_ENVIRONMENT_SET_DISK_CONTROL_INTERFACE, &disk_interface);
 
+   // Savestates
+   static uint32_t quirks = RETRO_SERIALIZATION_QUIRK_INCOMPLETE | RETRO_SERIALIZATION_QUIRK_MUST_INITIALIZE | RETRO_SERIALIZATION_QUIRK_CORE_VARIABLE_SIZE;
+   environ_cb(RETRO_ENVIRONMENT_SET_SERIALIZATION_QUIRKS, &quirks);
+
 	// Inputs
 #define RETRO_DESCRIPTOR_BLOCK( _user )                                            \
    { _user, RETRO_DEVICE_JOYPAD, 0, RETRO_DEVICE_ID_JOYPAD_A, "A / 2nd fire / Blue" },\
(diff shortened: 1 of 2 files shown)


MajorWilly:
Wow, thanks ! So it's fully working apart from rewinding ? I've never used the rewind feature in Retroarch so I personally wouldn't have noticed.

Is there anything left to add to this core (apart from updating UAE) ? It's come a long long way :smile:

sonninnos:
Works fine in my tests so far, with ADF/IPF/HDF & A500/A1200.

There are only 2 things left in my todo-list:
- Directory filesystem mounts
- CD32

And these are marked as done:
* No savestates
* Interlaced mode is fucked
* Option to enable NTSC aspect in PAL
* No CD32 controller
* Disable analog stick mouse if analog stick is keymapped
* VKBD is painful
* Not able to use joystick with keyboard buttons
* No keyboard controller
* Keyboard layout very fucked
* No reset
* Sound interpolation is fucked
* Sound core options not working
* No hotkey for UAE statusbar toggle & There should be only one statusbar
* No mouse in "use analog" core option
* GUI options should all be moved to core options
* Remove the whole GUI

MajorWilly:
Is the 'directory mounting' for WHDLoad games ? I've heard that a lot of games won't extract properly on anything but the Amiga File System (illegal characters).

One thing i've noticed, if you are playing a WHDLoad CD32 game and have the controls set to CD32 pad, you can't press a mapped quit key, or bring up the VKeyboard, Statusbar etc. You have to set the controls back to Retropad and then the mapped keys work again. I think I tried to explain this once before (but may have dreamt it!). Not much of a problem though.

sonninnos:
I use filesystem mounts with WinUAE and haven't yet encoutered a WHDLoad game that does not work.

Enabling the bonus RetroPad keys to the CD32 pad would be a nuisance, since there is only L2, R2 and Select left to use. So one would have to change the statusbar and vkbd keys to one of those. Possible sure, so maybe some time.

MajorWilly:
I see your point about the CD32 pad. It's not worth the hassle with only a few spare keys. It's easier to manually change the control type.

Just out of interest, how are you going to be using the directory mounts ? Will it be a Boot HDF or directory as DH0: and then the directory of games or single game data only HDF as DH1: ?

sonninnos:
I had a simple idea of just adding the unexisting buttons for CD32 pad as separate mapping options, so the RetroPad mappings won't matter. Analog sticks would stay the same to minimize the amount of more core options.

Yes, I use a small boot hdf as DH0: and the games directory as DH1: and a temp directory as DH2:. That way I can use the emulation most closely to the way I use the real thing. Having the middle-man obstacle of always making a separate hardfile or copying the data into a hardfile for each and every game, even when just for testing out if the game is worth playing, is something I want to avoid to the last breath.

But a simple and efficient script for that job gets my attention for sure.

The way I handle starting a certain WHDLoad game now through EmulationStation is that I create a file in the temp directory at launch named whdboot which just includes the name of the wanted directory. Then the Amiga is fired up without parameters. Startup-sequence checks if the file exists, reads it, chdirs to it and runs a scriptfile named "load". I create the script for every game with `echo whdload <game>.slave preload > load`. I don't mainly use Workbench, only Zshell (which unluckily will crash if whdload is run inside it, so the script needs to be run with `execute load`).

One could also start the .info with `wbrun <game>` (without .info) just like mouseclicking the icon does. Then all the custom parameters/tooltypes would have to be set through editing the icon, and that requires using the mouse.

Sure one might think that if there is so many hoops already in the process, why would a few more matter. Well.. attaching hardfiles in and out of WinUAE requires constantly bringing up the GUI. The best compromise would be to have one single hdf for the games. But without a filesystem temp mount the outside launch of a certain game is impossible. The implementation right now allows reading and showing directories, but any read from any file will crash.

Core options can't really help in that regard, except for the whdload-function it already has. One must use a custom uae config for the job, and that will again bring other problems like the name of the savestate would need to be renamed and handled outside of RA (like I did with SDLVICE and its quicksave feature).

Sorry for the wall of text, I couldn't make it any tighter, heh.

MajorWilly:
Here's a link to this guys script that creates single HDF's of each game straight from Retroplay's constantly updated LHA's (I assume you know of Retroplay that took over KillerGorrilla's WHDLoad packs ?).

http://eab.abime.net/showthread.php?t=48478

The script is quite complex. The result currently uses a boot HDF or directory (just add kickstarts) in DH0: and the game HDF in DH1:. It also includes a simple boot menu where you can choose which slave file to boot or read the readme/instructions etc. 

Maybe with your skills you could adapt the script to add the bootfiles to each HDF and thus making them standalone single bootable HDF's.

sonninnos:
Thanks, I shall investigate.

MajorWilly:
Did you have any luck with altering that script or did you not bother yet ? :smile: I just thought single bootable HDF's would be good for RetroArch playlists.

sonninnos:
Haven't yet gone deeper than a glimpse. Other stuff on the way.

MajorWilly:
Please forget everything I asked about the WHDLoad script above. I've found out that it's already possible to use the method - Boot HDF (DH0:) + Game HDF (DH1:) by using the Core Option `Use WHDLoad.hdf`. It wasn't very clear, but after a quick experiment it worked. Thanks.

PR #128, opened 2019-09-22
Controller related updates

sonninnos:
Had to bring some dynamite, because there was something wrong with the controls. Now it feels pretty much perfect. Please prove me wrong.

Changed the keyboard handling so the underlying keyboard key isn't pressed if using keyboard as RetroPad. For example in Turrican II the CTRL key pauses, and if CTRL key was the fire button, you're out of luck. And in Banshee the CTRL key also does the flip, so CTRL key as fire was no bueno. It can be overriden with the same Right Alt which overrides cursor keys without changing device type to keyboard.

Also changed the behavior of the D-Pad mouse speed toggler. Now it acts as a modifier which goes either faster or slower depending on the new core option starting point.

And again a little less compilation warnings.

And the bonus: Included joyport indicators to the statusbar, as seen in the VICE core.

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -40,7 +40,7 @@ void gettimeofday (struct timeval *tv, void *blah)
 unsigned short int bmp[1024*1024];
 unsigned short int savebmp[1024*1024];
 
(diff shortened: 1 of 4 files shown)


rsn8887:
Awesome, thank you!!!!

PR #131, opened 2019-09-25
Proper dupefire location + Keyrah handling correction

sonninnos:
Fixed the duplicate button skip to affect only RetroPad and made Keyrah keys also act properly with the same axis shenanigans.

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -383,59 +383,73 @@ void ProcessKeyrah()
    /*** Port 2 ***/
 
    /* Up / Down */
-   if ( input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP8) &&
-       !input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP2))
-      {
-         setjoystickstate(0, 1, -1, 1);
-         kjflag[0][RETRO_DEVICE_ID_JOYPAD_UP]=1;
-      }
+   if (input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP8)
+   && !input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP2))
+   {
+      setjoystickstate(0, 1, -1, 1);
+      kjflag[0][RETRO_DEVICE_ID_JOYPAD_UP]=1;
+   }
    else
-   if ( input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP2) &&
-       !input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP8))
-      {
-         setjoystickstate(0, 1, 1, 1);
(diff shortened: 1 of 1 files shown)


MajorWilly:
I'm not sure if this commit fixes it (buildbots not ready yet) but in the statusbar when you change from joystick to mouse, there is something written after 'mouse' that overlaps the word 'zoom'. I think it says the same thing as the zoom level, small, medium, large etc.

P.S. The auto centering on the zoom is awesome ! The tooltip help thing says for best results turn off vertical centering. Is there a reason for this as i've left the centering both on 'simple' and seems to work better than 'vertical off' ?

sonninnos:
This does not change the statusbar in any way, but I can't replicate any overlapping with any settings.. Please take a screenshot. There is only supposed to be "Mouse:X" with X being the first letter of the speed setting.

The UAE's vertical centering does not really do much unless crop is used, other than a small placement shift when VKBD/statusbar is toggled.

And there seems to be a small problem with the sensing of the autocentering. As in it does work when triggered manually after the actual playfield is ready & visible, but it won't always center after cracktros.

MajorWilly:
Screenshot_2019-09-25-21-51-16

rsn8887:
It looks like that "None" shouldn't be there.

sonninnos:
How in the bleep is that possible..

Can't find a problem with analog mouse either.

rsn8887:
I wonder if this is really allowed ```Draw_text32((uint32_t *)bmp,STAT_DECX+160,STAT_BASEY,0xffffff,0x0000,FONT_WIDTH,FONT_HEIGHT,20,((MOUSEMODE==-1) ? "Joystick" : "Mouse:%s "), PASSTR);```

You are always passing the arg PASSTR even though "Joystick" has no `%s` in it. I am just guessing here.

sonninnos:
I'll try to move those out of there, but can't confirm if it works or not.

rsn8887:
I think I fixed it in this commit https://github.com/libretro/libretro-uae/commit/64f78810dbddad011721d0928251fc5ff9b08d7e

MajorWilly:
I think analogue mouse maybe OK. Because it was moving so fast I couldn't see any gradual speed changes. I put it on the lowest sensitivity and can see some slight analogue-ness happening now. :smile:

sonninnos:
I've tested only with Xinput. X360 and DS3 pads behave just the same.

And does any other core with analog stick control work properly? I'll see if the values are used in some other way in other cores.

rsn8887:
Maybe you have analog-to-digital enabled and are somehow using the digital mouse? Or you have the Retroarch Settings->Input->Analog Sensitivity not set to `1.0`?

MajorWilly:
The analogue is OK. I just think it was because the core options sensitivity was 100%. I probably won't use it as you have to change to mouse mode anyway to enable the mouse buttons so may as well just use the dpad. It's all good.

sonninnos:
The mouse buttons have been mappable to any RetroPad button for a while now.

PR #133, opened 2019-09-30
Improved zoom autocenter + analog stick mouse

sonninnos:
- Made zoom autocentering more reliable
- Removed the need for an ugly hack caused by PAL/NTSC difference in minfirstline
- Took influence from the Dosbox core for analog stick mouse control and tried to make it non-linear

Surprisingly even Cannon Fodder is pretty playable now with the stick.


rsn8887:
Thanks! Merging...

PR #134, opened 2019-09-30
Replaced "cpu speed" with "cpu throttle"

sonninnos:
Useless "real" vs "max" is now "-90%" to "+200%". Makes a rather dramatic difference in Indianapolis 500, for example. Maximum can be increased more if +200% is not enough.

Also reordered some related core options and made cpu options come to effect without restarting.

The bonus is a ninja option `waiting_blits` included in the opposite of `immediate_blits` since they cancel each other out. Fixes graphics corruption in Nebulus 2 on A1200, for example. No ill effects found yet at "automatic".

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -336,20 +336,20 @@ void retro_set_environment(retro_environment_t cb)
          "enabled"
       },
       {
-         "puae_cpu_speed",
-         "CPU speed",
-         "Needs restart. Max requires Cycle exact: off",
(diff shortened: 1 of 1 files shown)


rsn8887:
Thanks merging. Once we understand how to fix cpu_speed we might bring it back, but currently, yeah, it just doesn't work right.

MajorWilly:
Holy sh!t Batman ! :tada: :smiley:

PR #136, opened 2019-10-01
Combined cpu compatibility core option

sonninnos:
Less is more.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -338,31 +338,21 @@ void retro_set_environment(retro_environment_t cb)
          "enabled"
       },
       {
-         "puae_cycle_exact",
-         "Cycle exact",
-         "Combined CPU + Blitter cycle exact",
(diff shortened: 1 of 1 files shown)


MajorWilly:
Guys, just out of interest, do you know of any games that actually need the 'CPU Compatible' option ON ? So far i've found that having 'CPU Compatible' ON is actually less compatible ! IK+ and Jimmy Whites Snooker have missing/bad sound when the option is on. I always have it turned off.

sonninnos:
No idea, but the only way to confirm it is to keep the option available. I won't mind removing it if the only usable options are in fact normal & cycle exact.

sonninnos:
And according to the author, more compatible adds "prefetch pipeline", and

> Prefetch is needed by many copy protections and self-modifying code tricks.

So maybe the effect won't be seen that much with cracked ADFs..?

PR #142, opened 2019-10-07
Control improvements

### LINKED ISSUE #139: Having the ability to customize more controllers with the RetroPad settings in the Options menu
I'm quite new to this emulator core; I've been an avid user of both WinUAE and FS-UAE till now.

In some games, I like to map custom keys from the Amiga keyboard itself to the Xbox One controller. One example is in the game Blues Brothers, where F7 is used to switch between both of the two players in 2 Player mode. I personally prefer to bind that key to the Y-button on my Xbox One gamepads (RetroPad X in the Options menu of this libretro core).

However, it seems like it's currently only possibly to do that for the Player 1 controller in the Options menu; I have not figured out to assign the Y-button of my secondary Xbox One controller to F7 as of now.

Could it be an idea to put custom mappings for a secondary gamepad as well in the Options menu? The menu could maybe be a bit on the "longer side" afterwards, but it would be nice for 2 player games on the Amiga let's say if you both want to assign the Start button as the P button (Pause in many games).

sonninnos:
- Enabled shared RetroPad mappings for joyport 2
Closes #139 

- Prevented moving VKBD cursor while key is pressed

- Revised and finetuned emulated mouse controls
-- Mouse speed modifier replaced with 2 modifiers: slower & faster
-- Modifiers are directly mappable without the need of a keyboard hotkey, just like mouse buttons
-- Applies to both D-Pad and Analog stick mouse

The bonus:
- Removed unneeded fast forward trick
-- Seems to fix random drive sound volume changes after fast forwarding, and was useless anyway thanks to another related fix


rsn8887:
Awesome thanks!

PR #144, opened 2019-10-08
Mouse support for both normal joyports

### LINKED ISSUE #140: Multiple mouses in the 2 player mode in Lemmings 1
I'm not sure if this is anywhere near a highly featured request, but I'll put it here anyway.

I just did a test run with Lemmings 1 to see if the 2 player mode works for both players. Both my wired and my wireless mouse (for my HTPC setup) controlled the first player, and the right analog stick of the first Xbox One gamepad reacted only to player 1 during my short test.

Player 2 however seemed during my short test not possible to control anyway.

sonninnos:
Note that this does not enable the use of 2 physical mice, but only RetroPad emulated.

And since the mouse and joystick ports are reversed in Amiga, as in the first mouse port is the second joystick, and the second mouse is the first joystick, this simply works just like RetroPad works now: Both ports can control the other port as mouse directly via analog sticks in joystick mode, and mouse mode forces D-Pad and the first 2 buttons to mouse.

Hotkey mouse buttons and mouse speed modifiers apply to the controller in question.

There is very little leg room to make this any simpler and clear-cut without creating another device type, which would skip joystick control completely. I don't think that is necessary.

Bonus: A global mouse speed core option which affects all mouse types

Closes #140

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -58,9 +58,7 @@ int analog_right[2];
 extern unsigned int opt_analogmouse_deadzone;
 extern float opt_analogmouse_speed;
 extern unsigned int opt_dpadmouse_speed;
-int analogmouse_speed=0;
-int dpadmouse_speed=4;
-int fmousex,fmousey; // emu mouse
+unsigned int mouse_speed[2]={0};
 int slowdown=0;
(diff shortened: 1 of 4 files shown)


rsn8887:
Awesome, thanks!

PR #146, opened 2019-10-09
Support for multiple physical mice

sonninnos:
RA input driver needs to be set to raw in Windows and udev in Linux, afaik. Mouse index needs to be set correctly in RA input configs. Only tested with Windows.

Also added middle mouse button as a bonus.


sonninnos:
Had to add the core option. Otherwise both mice will be controlled by the same mouse, because RA has mouse index at 0 on all user inputs by default.

There is also something seriously wrong with raw input on Windows, because pressing enter closes or crashes RA.. Keyrah keypad will not work for player 2 either.

rsn8887:
Nice, merging! Maybe the raw driver in the Retroarch repository can be easily fixed with a PR? I bet it wasn’t tested much.

PR #151, opened 2019-10-12
Controller fixes

### LINKED ISSUE #150: The very first time I press the left- or the right direction in-game, the character moves without stopping
I've encountered a strange bug in this libretro core that's very easy to replicate; the first time you start any Amiga 2D platform game, are in-game, and press either the left- or the right directional button, the character walks in that direction without stopping. I'm using the wireless Xbox One gamepad (the one with bluetooth, although I'm using the wireless gaming receiver).

If you take James Pond 2, press the right directional button a single time in the very start outside the castle, and the character will walk through the first door, and he will not stop until he reached the invisible barrier on the far right on the first level. If you  then press the left directional button, the character will walk as far left as he can without stopping. Pressing left- or right afterwards and the directional buttons on the gamepad will behave as they should.

sonninnos:
- Fixed joystick port from going into mouse mode without moving it as a mouse
-- Caused left/right to stay down the first time they are pressed (but not up/down)
-- Came about in the multiple mouse update
--- This will also happen with the mouse port if used as a joystick, and it can't be prevented without changing the default mode to joystick, but that would make mouse to not work unless mouse button is pressed, so I think we can live with it
--- It could be fixed though with a dirty hack of just pressing left and right twice silently at startup, and it would not affect regular mouse usage
- Fixed broken Keyrah handling caused by a typo
- Cleaned up key processing functions

Closes #150

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -436,7 +436,7 @@ void ProcessKeyrah()
    if (!input_state_cb(0, RETRO_DEVICE_KEYBOARD, 0, RETROK_KP6)
    && kjflag[0][RETRO_DEVICE_ID_JOYPAD_RIGHT]==1)
    {
-      setjoystickstate(0, 1, 0, 1);
+      setjoystickstate(0, 0, 0, 1);
       kjflag[0][RETRO_DEVICE_ID_JOYPAD_RIGHT]=0;
    }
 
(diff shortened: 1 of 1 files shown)


sonninnos:
Shall I add that above mentioned hacky thingy too?

rsn8887:
I think this is great as is. Thanks! Merging.

PR #154, opened 2019-10-15
Controller fix + legacy removals

### LINKED ISSUE #152: Player 2 character walks forever to the left- or right direction on the first button press (after the Controller Fixes #151 Pull Request)
I just got the 67fc6c6 update of this libretro core from the Online Updater in RetroArch, and I tested it out. The Issue #150 that I wrote in the Issue tracker is mostly fixed during my quick testing, and by mostly it seems like Player 1 right- and left directional keys behaves OK now.

The Player 2 gamepad however acts like it just recently did; the first single press in the left- or right directional button causes the character to run all the way in that direction (tested with Blues Brothers and RodLand).

sonninnos:
- Joyports should now obey the appropriate controller type without jiggle initiation if joyport is unused before that
Closes #152 

- Removed obsolete core options, because zoom is über alles:
-- Crop overscan
-- Horizontal/Vertical centering

- Removed configuration files to not give the impression that they are required to fiddle with anymore
- Out-of-date EUAE configuration.txt replaced with a more suitable PUAE one from version 2.8.1b6 (rest of the source should be updated too)

- Combined statusbar core options to one core option
- Reorganized core options to be more section like
- Tidyings and reindentations


rsn8887:
Thanks! Merging...

PR #156, opened 2019-10-17
Vertical positioning + Autozoom + 2nd fire improvements

sonninnos:
- Vertical offset manipulation changed from minfirstline to thisframe_y_adjust
-- Fixes drawn line calculations = more reliable autozooms and autocenterings
-- Also fixes interlaced modes statusbar position shifting, like Wizkid

- Changed 2nd fire pull-up resistor behavior, fixing 2nd fire working only once on certain games, like Banshee

As always, testing and reports are more than welcome!

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -75,12 +75,12 @@ extern int prefs_changed;
 int opt_vertical_offset = 0;
 bool opt_vertical_offset_auto = true;
 extern int minfirstline;
-static int minfirstline_old = -1;
 extern int thisframe_first_drawn_line;
 static int thisframe_first_drawn_line_old = -1;
 extern int thisframe_last_drawn_line;
-static int minfirstline_default = VBLANK_ENDLINE_PAL;
-static int minfirstline_update_frame_timer = 3;
+extern int thisframe_y_adjust;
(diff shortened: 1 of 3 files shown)


rsn8887:
Thanks!

MajorWilly:
I tried about 3 games that have 2 button support and they all worked great. I wasn't expecting this to be fixed because I think you said it also happened on WinUAE. 

The Autozoom is absolutely awesome. Reduces the need to keep pressing the zoom button to find the best fit for each game. I know it won't be perfect, on a quick test I found the 'Bubba'n'Stix' title screen was the wrong zoom level but can still be corrected with the Zoom hotkey. It's to be expected. I imagine Demos would make the Autozoom go crazy !

You never cease to amaze.

sonninnos:
Cheers! I decided to take a jab at the 2nd fire after I noticed it worked fine when both of the joyports were accidentally in mouseport mode for a while after the multiple mouse update, and naturally the issue came back after fixing it. The solution was extremely simple after that revelation.

And let's hope my spider sense tingles right on that autozoom issue too. I'll look at it soonish.

sonninnos:
The hunch was correct, fix coming soon.

PR #157, opened 2019-10-17
Autocenter optimization

sonninnos:
Better workaround for the constantly running timer (which makes sure screen is centered and zoomed correctly), because there is no reason for it anymore, other than when a savestate is loaded.


rsn8887:
Thank you!

PR #159, opened 2019-10-18
Autocenter improvement

sonninnos:
- Triggers geometry update also when last_drawn_line changes
-- Fixes autozoom in Bubba n Stix, for example


rsn8887:
Awesome! Thanks.

PR #163, opened 2019-10-19
Autozoom calculation correction

sonninnos:
Just a small fix for one missing line. Noticeable in Cannon Fodder intro logo screen (missing the white line at the bottom) and gameplay (missing last line from the map globe), for example.


rsn8887:
Thanks!!!

PR #165, opened 2019-10-25
VKBD corrections and tidyings

sonninnos:
Some minor maintenance and cleanups.

- VKBD had a few buttons labeled wrong according to US keymap
- The keyboard size now stays pretty much the same with both PAL and NTSC when toggled without changing timing
- Removed distracting key borders in the poor man's "transparency" mode


rsn8887:
Looks good! Thanks! Merging.

PR #166, opened 2019-10-29
Directory filesystem fixes

sonninnos:
Reading and writing with WHDLoads seems to be finally ok now. Only confirmed on Windows, so let's not update Readme just yet.

- Used/available size faked like in FS-UAE
- Protect flags on all files are +e, so executing works without Workbench and/or explicit flagging
- "." and ".." are for some reason seen as files and not directories, so enabled hiding them too


rsn8887:
Thanks!

PR #167, opened 2019-10-31
Internal floppy sounds

### LINKED ISSUE #161: Floppy sound emulation doesn't seem to work
It seems like the slider for floppy sound doesn't do anything at all as of now.

sonninnos:
No more need to hassle with external floppy sounds unless one wants to. The internal sounds are the same default sounds as before, but with the volume turned down a notch. Kept only A500 and LOUD, since A500Real was the same files as A500 and who would want ST here anyway.. Some files are edited slightly for better consistency. The source is `drive_sounds.zip` from winuae.net.

Also changed the default CPU compatibility option to "Cycle exact", since there seems to be more problems without it, especially with savestates.

Closes #161


rsn8887:
Thanks!

PR #171, opened 2019-11-14
Bunch of minor changes

sonninnos:
For yet another reason (more ways to skip cracktros) I decided to map the mouse to the VKBD. It would be better to have totally separate buttons, but for the time being I jammed them in the other page, because I could not figure out what buttons to fill a whole row with. Maybe reset and turbo fire toggle like in VICE.

It does not make much sense, but with the cycle-exact speed option at 1x the WHDLoad version of Flood is suddenly at the correct speed. Too bad it does not help with PushOver. And so far overclocking does not seem to have any benefit except in SysInfo benchmark.

Maybe I should ditch the statusbar zoom indicator completely, because that would help with the lores mode not needing two rows.


libretro/libretro-mapper.c
```diff
@@ -1170,14 +1170,16 @@ void update_input(int disable_physical_cursor_keys)
 
 void retro_poll_event()
 {
-   /* If RetroPad is controlled with keyboard keys, then prevent up/down/left/right/fire/fire2 from generating keyboard key presses */
+   /* If RetroPad is controlled with keyboard keys, then prevent up/down/left/right/b/a/x/y from generating keyboard key presses */
    if (uae_devices[0] == RETRO_DEVICE_JOYPAD && ALTON==-1 && 
       (input_state_cb(0, RETRO_DEVICE_JOYPAD, 0, RETRO_DEVICE_ID_JOYPAD_B) ||
-       input_state_cb(0, RETRO_DEVICE_JOYPAD, 0, RETRO_DEVICE_ID_JOYPAD_A)
+       input_state_cb(0, RETRO_DEVICE_JOYPAD, 0, RETRO_DEVICE_ID_JOYPAD_A) ||
```
rsn8887: Doesn't this prevent e.g. mapping a keyboard key to button X?

rsn8887: What I mean is: "Doesn't this prevent me from using mapper to map button X to act as a keyboard key"?

I am probably wrong.

rsn8887: I am wrong.

rsn8887:
Thanks.

libretro/libretro-mapper.c
sonninnos: Aye, f.ex if RetroPad X is pressing Space via mapper and Space is RetroPad X, with pass-through enabled Space will be pressed twice, and now only once with pass-through disabled.

PR #172, opened 2019-11-17
Miscellaneous bunch

### LINKED ISSUE #164: Horizontal Position
Same as Vertical Position only for horizontal.
Games like north and south are left aligned for which such a feature would be perfect.

sonninnos:
- Statusbar simplified to fit in one row in lores mode
- Zoom mode hotkey toggles the mode on/off and not jump to the next one
- Horizontal position core option with automatic centering
-- Does not throw Banshee off screen, and does keep Turrican centered, but for some reason makes North & South less centered..

Closes #164

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -80,6 +80,7 @@ extern int mapper_keys[32];
 extern int video_config;
 extern int video_config_aspect;
 extern int zoom_mode_id;
+static int opt_zoom_mode_id;
 extern bool request_update_av_info;
 extern bool opt_enhanced_statusbar;
 extern int opt_statusbar_position;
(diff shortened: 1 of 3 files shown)


rsn8887:
Thank you!

MajorWilly:
@sonninnos It's been a while! I noticed that the new 'Horizontal Automatic' option displays some things wrong like 'Big Time Sensuality' demo. I'm not really sure what the point of this option is ? Turrican and Banshee are centered correctly without 'Horizontal Auto' (They've been correct since the Auto Zoom and Overscan omission).

sonninnos:
It does the same as the internal horizontal centering 'smart', but without the twitching on vkbd/statusbar toggle, plus it allows manual adjusting. Vertical centering only comes to play with zoom modes, but this horizontal does not care about zoom.

Turrican is not exactly centered.

Setting 0, as in default:
image

Automatic, as in smart:
image

It can also been seen in action for example with Workbench's Preferences application by moving the screen around.

MajorWilly:
Not sure why Horizontal Auto would affect this ? Big Time Sensuality demo -

Horizontal Correction Off -
Screenshot_2019-11-19-16-34-42

Auto -
Screenshot_2019-11-19-16-34-18

sonninnos:
Looks like it is doing some zooming when it shouldn't.. I'll look into it.

PR #177, opened 2019-11-22
Minor faults eliminated

sonninnos:
- Zoom toggling fixed to obey core option better
- No more crashing when ejecting a non-existing disk, thanks to an unwarranted strlen
- Statusline made shorter by replacing last floppy LED with hard disk LED when HD enabled
- Removed statusline floppy indicator track numbers when ejected and also HD text
- More readable and useful logging at startup and at runtime


rsn8887:
Thank you! Your PRs are always making me happy.

PR #178, opened 2019-11-22
File support changes

sonninnos:
- DMS support was not working at all despite being on the valid extension list
- Internal ZIP support is not working (and did not succeed in it), and a ZIP will never go that far in the first place, because RA does the unzipping, so changed it to ADZ (gzipped ADF) which does work

Also is there a real reason to keep the FDI format? I tried to search for a disk for testing but couldn't find one in a reasonable time.


rsn8887:
Thank you!

PR #181, opened 2019-11-30
Basic maintenance

sonninnos:
- Altered zoom mode centering value limits
- Added the rest of the RetroPad buttons to keyboard pass-through option
- Added core option for showing/hiding mapping options


rsn8887:
Thanks!

PR #183, opened 2019-12-01
Automatic default modes for core options: model & filter type 

sonninnos:
Since the default A500 is not HD bootable, the new default automatic mode makes juggling between floppies and hard drives more convenient. It selects A500 with floppies and A600 with hard drives, if there is no filename override tags.

Also made the new automatic sound filter type the default, because of course it should match the used model.

Removed the need for ()-chars with OCS, ECS & AGA tags, because AGA games tend to already have AGA in the title, so no need to make it `AGA (AGA)` or change it to `(AGA)`.

Altered A500+ model to be more realistic with only chip ram.

And the bonus is yet another tinkering with the keyboard pass-through.


rsn8887:
Thanks!

PR #184, opened 2019-12-03
New WHDLoad.hdf + overriding core option

### LINKED ISSUE #179: Having the option to enter custom commands & run trainers in WHDLoad hdf-files?
Being a former user of fs-uae before fully migrating into this libretro core for my Amiga emulation needs, there is one thing I'm missing from fs-uae that as far as I know is missing in libretro uae:

The options of entering custom commands and use the bundled trainers that comes with some of the WHDLoad Slaves to some of the Amiga games.

For instance, The Settlers has the ability to skip the intro by entering Custom1 as a launch-command before starting the WHDLoad. Some other games have optional trainers at the very start of the games, in other words when the AmigaDOS screen is in the background and the tiny box in the middle has a description with the name of the game you're running, before the actual game starts (in this box the trainers are often placed by having various toggle-able checkboxes, and info like type Custom1 to skip intro is displayed here).

There are however only a handful of games where I personally miss the introduction screen before the game actually starts (because of trainers and relevant info), other than that I think it's fine the way it is now. Maybe having a toggleable option for this introduction screen in this core's Options menu?

sonninnos:
Crucial improvements added to README.md, so no need to list them here. Nothing should prevent the old HDF from working like it does now.

Please report if I missed or overlooked something! Almost pushed the wrong hdf version which had the kicks and my registration key included, heh.

Closes #179


rsn8887:
Looks good! Thanks.

PR #188, opened 2019-12-21
Analog stick deadzone calculation + Disk change force

### LINKED ISSUE #186: Flashback - Switching disks doesn't work
I understand there's a separate core for this game specifically, but when I try to play this game, it runs well right up to the point where it says to insert disk 2 after the title screen. No matter what I try, it simply just doesn't work - it flashes black for a second and then says insert disk 2 again. Tried pressing every button and nothing happens. On WinUAE for PC, switching to disk 2 causes the screen to go black for a while and then goes to the first cutscene as it should. Other multi-disk games I've tried work. What's the problem here?

sonninnos:
- Brutish analog stick deadzone calculation changed to slick vector magnitude
- Fixed disk change forcing

Whoever did the `DISK_check_change()` hoping it would do what it says it might do probably didn't look what it actually does. Those calls will be removed, since the function will be run anyway, making them being there pointless, and since the function itself only:
- checks if floppy speed has changed
- checks if floppy type has changed

Both unlikely to happen during normal use and neither play a part with the actual disk operation at hand. Replaced with `DISK_reinsert()`.

Closes #186 

Bonus:
- Statusbar shows CD32 Pad button colors


rsn8887:
Thanks! Merging.

PR #190, opened 2019-12-25
Video maintenance

sonninnos:
New features:
- Video resolution core option no longer requires core restart, but only a soft reboot
- PAL resolution changed to full 576p/288p
  - A minor last line drawing thing preventing the use before is now dealt with

Issues solved:
- Many interlaced screens got zoom calculations off the rails and so zoom shaked horribly, like CD32 boot screen and Turrican Anthology Compilaton start menu
- Fire & Ice CD32 not triggering zoom correctly and not centered horizontally
- Crashing with 24-bit hires double line mode

Bonus:
- Combined CD32 KS + extended ROM file support option with the same name
  - CD32 extended ROM only searched if KS ROM size is 512KB, combined will be 1MB
- Cleanups


rsn8887:
Thanks!

PR #198, opened 2020-01-02
File format support bonuses

### LINKED ISSUE #19: Autoconfig and launching of whdload games, archives.
fs-uae has a very nice feature that makes playing amiga games much cleaner. They use this database:
https://openretro.org/about

to collect configs of checksummed archives (including WHDLoad archives + the actual game files) and create a tmp hd, use one the workbench utilities to copy over the files to it and start the game just like that. Modified files at the end are saved. You can check out the process on the fs-uae github. Needlessly to say, i think 0-configuration game playing would fit in with the RA goals for this complex computer system, considering that a great deal of options on the amiga configs and a very many user error possibilities exist.

WHDLoad uses lha archives to preserve amiga filesystem bits for playing on a real amiga. Obviously, fs-uae is able to ignore that requirement (although it appears to use a workbench program to do something to the files, i think whdload patching). fs-uae accesses that data remotely (i had to register to openretro).

One little subtlety is that a great deal many WHDLoads around the net are corrupt in a strange way. The files inside have corrupted names when extracted from lha: instead of recreating the directory structure, they replace it with a flat view of the files with the parent directories names prefixed to the file name with '%5f' serving as dir separator. I'm unsure if this is a 'real amiga' thing or not.

Needlessly to say, fs-uae won't load these correct when trying to load such a archive. This is very often not noticed because fs-uae-launcher does something to fix these if you happen to be registered on openretro: it compares the sha256 checksum of the expected files to the checksum of the archive files and if they all match, extracts the files to the right place and name (taken from the database), thereby 'fixing' the extraction.

I'm unsure if this is a good idea to continue or not. On one hand, roms are dependent on a online service to function correctly, on the other, roms run. Maybe a one time 'fixing' step would be better, maybe as a external app. OTOH if indeed this is a 'amiga thing' needed for correct operation of lha on the amiga and the files are 'correct' the extraction process needs to be made less buggy so results of extraction are compatible.

### LINKED ISSUE #48: Support for open-retro database for autoconfiguration
[FS-UAE launcher](https://fs-uae.net/launcher) uses the [open retro game database](https://openretro.org/) to auto config and boot several kinds of dumps (including whdload).

This program 'solves' many things that make playing amiga games extremely annoying, including model and settings, but also 'automatically create a hd image for whdloads' and 'cache that copy so you can keep the image without modifying the original' and is also planning to catch the whdload quit event to send the specific whdload quit command (it needs this to write buffers to disk because for speed, whdload usually only writes to file on quit).

I'd love something like this in retroarch (along with a continuously updated whdload hash database).

### LINKED ISSUE #60: Make Retroplay set of WHDLoads in lha's work
Whdloads in this set are done to be extracted on a 'real amiga' and won't be correctly extracted on normal unzip utilities. The retroplay set is all lhas. Strangely there are some hdf images, but I *think* they're also compressed with lha, which makes sense if they are supposed to be extracted on a amiga. I'm not certain because there is no real way to verify the lha archive directory structure in linux nowadays.

tl;dr:

1. create a native lha extractor and parse the whdload.slave file inside the lha passed to the core, to figure out the final machine type and other exotic configs it needs (extra memory, cd32 controller added support, things like that). If it's already a hdf inside, extract that as the 'new file' and parse the slave from it, possibly by extracting it from inside emulation with another amiga program.
2. create a sufficiently big hdf image (somehow) to contain the new game if the final file doesn't exist in retroarch/saves yet goto 3, if it does, goto 4.
3. create a native mounter for the hdf file and extractor of lha that preserves amiga file properties and weird amiga names (this might be difficult because the amiga has different forbidden characters and file attributes, so the normal filesystem interfaces - mount, normal cp etc, might not serve) OR create a auto running amiga program that does the extraction into the hdf from inside emulation and run it.
4. finally mount whichever hdf resulted with the properties from step 1.

Then later a prs could be added for libretro-database so the scanner recognizes these lha's as the games they are, and kept uptodate periodically. As the entry points, they'd be read only and the 'real cached game' would be on retroarch/saves or similar, like certain other cores handle writable memory in 'ROMs' from early consoles.

Retroplay warns that it's difficult to make a extractor that works on the native side (because people apparently constantly break the installs trying with some 'standard' unix lha extractor):

https://forums.libretro.com/t/uae-core-for-retroarch/15599/46

So basically two paths; either create a specialized extractor that preserves the extra metadata and names from the amiga created lha (probably by not touching the 'modern' filesystems and using a library that can write hdf files and attributes directly from in-memory bytes) to output a valid amiga filesystem, or create amiga autoboot programs that extract lhas, create hdfs and somehow return to outside emulation the data in the whdload.slave for config, which is basically the same but with less filesystem 'fun' and more amiga 'fun'.

This library: https://github.com/lclevy/ADFlib (literally found by search 'amiga filesystem' on github)

which purports to be a "A free, portable and open implementation of the Amiga filesystem" and specifically says it supports 'UAE hardfiles' (hdf) may help for both.

sonninnos:
Added:
- LHA for WHDLoad hard drive images (Retroplay)
Closes #19 
Closes #48 
Closes #60 

- MDS for CD images

Fixed:
- Updated WHDLoad.hdf hopefully for the last time (for a while)
  - Ready for read-only drives such as LHA
- Directory filesystem reworked, influenced by FS-UAE
  - Replaced WIN32 specific droppings with the usual suspects, works better with Windows
  - Re-enabled the Kickstart copy part for Android+Switch to get more data. Please test!

Bonus:
- Cleanings


rsn8887:
Thanks!!!

MajorWilly:
I've only quickly checked on Android. It doesn't crash anymore when trying to copy the Kickstarts, it works OK. WHDLoad.prefs work. There are still a few errors with non-WHDload HDF's. I haven't tried LHA support yet as I need to create a playlist because Retroarch doesn't recognize them when using 'Load Content'. I'll try tomorrow.

MajorWilly:
I tried several LHA's and they all worked great! You've done an amazing job on this, thanks. And that Custom feature when you quit a game, genius. We can now do things like change the table on Pinball Prelude. Where is it saving the custom file ?

sonninnos:
Thanks-a-bunch!

The custom is saved to WHDSaves: and it is saved per slave like `custom_$SLAVE`. Just realized that all those who have renamed the slave to game.slave are screwed with this feature. I can't handle and support every bad decision to the end of times though..

MajorWilly:
Well there's no need to use those custom made game.slave files anymore. Time to move on. They can use the original and constantly updated LHA's from Retroplay.

MajorWilly:
I found one that doesn't work. If this is what I think it is then there will be more. `BattleIsle&DataDisks` says that the files are corrupted. Is the LHA support doing what FS-UAE and WinUAE does to keep the file attributes and special illegal characters ?

sonninnos:
With that `BattleIsle&DataDisks` I get the same 

> data/disk-files are damaged or it is an unsupported version

even with WinUAE no matter which method I use.

This internal LHA handling code appears to be removed from FS-UAE, so I guess it extracts the archive first. This is doing what WinUAE does, which is access the LHA directly.

MajorWilly:
Found another - `SuperStreetFighter2Turbo_v1.1_AGA_0735` (and the WHDLoad CD32 version). It says `Exception "Line 1111 Emulator" ($2C) at $4 occured`

They work on WinUAE when I use ClassicWB and mount the LHA so maybe the WHDLoad.hdf is missing required libraries or something ?

sonninnos:
Do try without the ClassicWB just like PUAE does. First harddrive is WHDLoad.hdf.

I suspect that kind of error relates to something else entirely, like the accuracy state of the emulator version. As in cycle-exact isn't exactly done here yet.

MajorWilly:
I tried using the WHDLoad.hdf in WinUAE (DH0:) and the LHA in DH1: but got an endless black screen.

I used ClassicWB again but with a standard A1200 config and no cycle-exact (matched PUAE settings) but they ran OK, no errors.

sonninnos:
The game image is supposed to be DH0: and WHDLoad.hdf is WHDLoad:

MajorWilly:
With BattleIsle I tried the WHDLoad.hdf as WHDLoad: and the Game LHA as DH1: It gave the same error  as PUAE `data/disk-files are damaged or it is an unsupported version` . I tried with and without Cycle Accuracy so something must be missing in WHDLoad.hdf.

sonninnos:
Hmm, if I do `wbrun BattleIsle` then it works.. That does the same as using the mouse in WB. Checking out the tooltypes reveals it also sets `Custom1=3`. Adding it with MkCustom equals tadaa. I'm not aware of a method to list out the tooltypes in CLI, so these sneaky slave creator tricks can not be sensed..

MajorWilly:
I wonder if FS-UAE can run this game (BattleIsle) directly because it doesn't use a Workbench to run them ? And I suppose SSF2Turbo is just Emulator Compatibility related ?

MajorWilly:
I see what you mean now with BattleIsle. A few of these WHDLoad games rely on the icons setting the Custom options. Custom1=1 is the intro, Custom1=3 game and Custom4 and 5 are the data disks.

sonninnos:
I'm afraid it is. SnoopDos revealed no files missing.

MajorWilly:
OK thanks. Seems it's another poor slave. I have a non-WHDLoad HDF that works fine. (Port is not that great though. CU Amiga magazine gave it 83% !).

PR #203, opened 2020-01-06
Basic maintenance

sonninnos:
- Fixed analog mouse stick properly
- Fixed statusbar OSD fonts
- Restored default blitter wait to automatic, thanks to Rick Dangerous WHDLoad
- Added `savestate_initsave()` for savestates to enable compression
  - Rewind won't crash anymore, but still isn't usable
- Updated WHDLoad.hdf
  - Autocopy KS 1.2 for Arcadia installs
  - Fixed xpkmaster.library
  - More compatibility with non-WHDLoad installs


MajorWilly:
So what did Rick Dangerous do that made you restore the Automatic Blitter ? I thought you removed this because there was a suspicion that it might be messing up savestates or something, or is all that a thing of the past ?

sonninnos:
The graphics are all messed up without it (Try A1200 setup with WHDLoad and Cycle-exact enabled).  The savestate issue only came up with Cycle-exact disabled, and since Cycle-exact is the default, so should this be the default, since this affects Cycle-exact.

WinUAE also enables automatic blitter waits with the A1200 quickstart config.

MajorWilly:
That's probably why I didn't notice it. I use A600 for WHDLoad ECS games to keep the speed correct. A1200 makes a lot of WHDLoad games play too fast unless you use Cycle Exact and that's too demanding for my device (and some still play the wrong speed). I've also noticed a few games don't have sound effects when using A1200 (New York Warriors). 

This must be a PUAE issue because I tried it on WinUAE A1200 with Cycle Exact OFF and Blitter OFF and it plays OK.

rsn8887:
Thanks!

MajorWilly:
Whereabouts does this happen on Rick Dangerous ? I tried on PUAE A1200 Cycle exact, wait for blitter on & off and didn't see any glitches.

sonninnos:
I had slave 1.2, and looks like it is fixed in 1.3:

> version 1.3 (12.02.15) done by Wepl:
	- highscore saving fixed, did save before name was copied to table,
	  because of this name was missing in saved data (issue #2923)
	- **some blitwaits added to fix graphics glitches on faster machines**
	  (issue #3190)
	- splash window config for trainer added
	- gfx distortion when enabling raster dma fixed
	- updated install script

MajorWilly:
Dude, that was 5 years ago! :smile:

TyroneSlothrop:
Hi I'm afraid that for MacOS this latest version crashes - I'm not purposefully trying to stick to the latest version by the way, the one issue I had on mac with the previous version is that the mouse capture/movement didn't work (WHDLoad was though btw). All I want to do is play Worms (issue with mouse) and SWOS 2020 (that was fine).

I do really appreciate the recent work being put into the core though, Mac has been missing a decent amiga emulator and retroarch centralises the experience well.

sonninnos:
Thanks for reporting. That function was my hunch also, so an update shall be coming soon. Too bad I can't confirm since I only have Windows.

sonninnos:
Please try updating the core now.

TyroneSlothrop:
Yes confirmed that works now! Amazing (p.s. my mouse issue is also solved using a controller and the analog sticks, the macbook trackpad doesn't seem to work for mouse capture but I suspect that is more of an issue with the 'coacoa' input driver in retroarch rather than this core).

Thank you so much!

sonninnos:
Excellent! If the mouse works in RA it should also work with this core.

PR #206, opened 2020-01-08
Add support for auto-generated save disks in M3U playlists

jdgleaver:
This PR adds the ability to automatically generate one or more blank save disks when creating an M3U playlist.

- On any line of the M3U file, the following my be entered: `#SAVEDISK:VolumeName`

- `VolumeName` is the disk label - this is optional, and may be omitted to create an unlabelled disk image (only certain games require save disks to have specific labels)

- Save disks are created in the user's save directory, with the following name: `[M3U_FILE_NAME].save[DISK_INDEX].adf`

@sonninnos I realise you are the maestro when it comes to this core - please let me know if you object to anything I've done here!

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,29 @@ Note: ZIP support is provided by RetroArch and is done before passing the game t
 
 Append "(MD)" as in "MultiDrive" to the M3U filename to insert each disk in a different drive for games that support multiple drives. Only possible if there are no more than 4 disks.
 
+For games that require a dedicated save disk, one may be generated automatically by entering the following line in an M3U file: `#SAVEDISK:VolumeName`. `VolumeName` is optional and may be omitted. For example, this will create a blank, unlabelled disk image at disk index 5:
+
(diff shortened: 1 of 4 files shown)


sonninnos:
Cheers for the helping hands! Suits me fine, except personally I'd rather use savestates instead of in-game floppy saves.

jdgleaver:
Ah, thanks for the feedback!

> except personally I'd rather use savestates instead of in-game floppy saves.

Completely understandable, but I grew up doing it the 'hard way', so it's part of the experience for me  :)

I'd also like to have a go at implementing a fs-uae-style 'floppy overlay' feature (https://fs-uae.net/floppy-images) for games that save to the actual data disks (this is currently a problem on Android, since you generally don't have write access to external sd cards, which is where I'd normally place rom files) - but this looks like it might be a little tricky, and I have some pressing frontend stuff to get on with first...

BTW - thank you so much for all your work on this core. It's absolutely incredible now - a real pleasure to use!

PR #208, opened 2020-01-09
WHDLoad.hdf update yet again

### LINKED ISSUE #207: Some WHDLoad LHA games crash the core (due to 0 byte files etc.)
These WHDLoad LHA games crash the core/Retroarch.

Cadaver&ThePayoff_v2.4_0900
CannonFodder2_v1.12_0104
CaptainBlood_v1.1_0443
DefenderOfTheCrown_v3.1_CDTV

I've noticed that all 4 of these games have a 0 byte file inside. WinUAE uses a UAEFSDB file for dealing with illegal characters/files etc. Is that what PUAE uses ?

http://eab.abime.net/showthread.php?t=5126

(Just copied these posts from the WHDLoad.hdf thread as it wasn't related). There will be more that crash but I stopped after it was confirmed what the cause was.

SimCity_v1.0_2193
Zool_v2.0_CD32

sonninnos:
- Better compatibility
- Hold down fire button at boot for `.info` selector
- Hold down spacebar at boot for `Readme` and `MkCustom`

Also with the essential fix for LHA archives, where empty directories would crash the core.
Closes #207 

And some ReadMe cleanups.


MajorWilly:
Glad you found the issue with the crashing as there are some classic games that were affected. :+1:

MajorWilly:
Glad you managed to fix the issue with the crashing as there are some classic games that were affected. :+1:

sonninnos:
And luckily the fix wasn't too time and nerve consuming to find.

PR #211, opened 2020-01-11
Sound updates mostly

sonninnos:
- Fixed stereo separation channel reversing
- Fixed stereo separation from mixing down also CD audio
- Added core option for CD audio volume
- Minor cleanups on aisle Statusbar


rsn8887:
Awesome!

PR #214, opened 2020-01-12
VKBD usability improvements

sonninnos:
- RetroPad Start = Return
- More accessible mouse control (removed from NUMPAD page)
- Extra function keys
- Fixed a 16-bit color goof manifested on previous update
- Another weird horizontal centering issue sorted

retroarch_2020_01_12_15_16_36_789


MajorWilly:
Just a suggestion, do you think 'Select' should be default for the Virtual Keyboard ? If you have CD32 Pad selected as controls then Start + L + R are unusable, but you can access `swap J/M` and `Status Bar` from the Virtual Keyboard (and a WHDLoad Quitkey).

What is `TRBF` and `ASPR` ?

sonninnos:
I'll have to ponder on that. The goal is to keep VICE and PUAE default controls as similar as possible.

TRBF = Turbo Fire Toggle
ASPR = Aspect Ratio Toggle

MajorWilly:
I just thought that if people don't bother to remap anything then if they change to CD32 controls the keyboard is still active to access the StatusBar. Some CD32 games use the mouse and if it's a WHDLoad CD32 game that requires the QuitKey to be pressed to save anything (or other keys).

sonninnos:
PAL/NTSC of course. Just shortcuts for the existing features.

PR #224, opened 2020-01-19
Add build-time option to run the emulator in a thread instead of libco

realnc:
Currently, the emulator is wrapped with libco. This has some negative implications for debugging. Possibly other issues as well when it comes to the emulator doing system calls, or when saving/restoring registers on modern CPUs when switching context.

This adds a build-time makefile flag (`USE_LIBCO`) to libco with a pthreads-based mechanism. By default it's set to 1. Setting it to 0 will use pthreads.

I did observe a performance decrease however compared to libco. It's a bit difficult to measure exactly due to FPS fluctuation when using RA's fast-forward feature, but it seems to be around 3% to 5% or so.


rsn8887:
Is there any benefit apart from debugging use? I’d rather avoid performance decreases just to allow better debugging.

realnc:
> Is there any benefit apart from debugging use? I’d rather avoid performance decreases just to allow better debugging.

Other than possible avoidance of weird bugs and being able to reason better about whether or not an issue is the fault of the code or libco, no. All I can say is that wrapping a piece of software in libco with that software not having been specifically written with libco in mind makes things hard to reason about.

To give a more concrete example, when writing code in dosbox it was sometimes necessary to clear, set or check for null in cases where this was clearly not necessary. Like the C++ `this` pointer becoming null, for example. Updating libco fixed it, but I've been very wary since then about wrapping anything in libco that wasn't actually designed to be wrapped in libco.

Even when there's a small perf decrease, my view is that you can't put a price on correctness. But that's just me.

Anyway, the CI reports an error for the wiiu build:

    undefined reference to pthread_cancel

If the wii u toolchain doesn't have `pthread_cancel` this might put an end to this PR anyway :-P

hizzlekizzle:
it could be valuable as a compile-time debugging option

realnc:
PR updated. Now pthreads is optional (build with `make USE_LIBCO=0` to disable libco).

jdgleaver:
This is such a nice and clean implementation.

It looks like it could be a runtime option (on platforms with pthread support) - requiring a restart, of course. Might that be feasible?

realnc:
> It looks like it could be a runtime option (on platforms with pthread support) - requiring a restart, of course. Might that be feasible?

I'll certainly look into it. But if there's no other issues with this PR, can it be merged now and I can add a runtime option later on?

jdgleaver:
Oh, that's not for me to say! rsn8887 is in charge of this repo, so he'll make any decisions.

I just wanted to compliment you on your fine implementation  :)

jdgleaver:
Having said that, you'll need to add `USE_LIBCO` entries to `Android.mk` (https://github.com/libretro/libretro-uae/blob/master/jni/Android.mk) before this can be merged.

Otherwise, I can see no objections!

realnc:
Right. Didn't even notice `Android.mk` since `Makefile` has an `ifeq ($(platform), android)` in it. Seemed like Android is handled there.

Fixed.

jdgleaver:
Thanks! Yes, almost all the cores switched over to the ndk-build JNI system quite some time ago.

realnc:
Thanks as well.

This thread/libco thing might go away altogether at some point though. It seems somebody integrated PUAE's main loop properly (https://github.com/hiddenasbestos/libretro-uae/tree/gridc). Haven't looked into it at all yet, but if it works correctly then we could switch to that at some point.

jdgleaver:
Oh wow - that's a good find! That could be a very neat improvement.

jdgleaver:
@realnc So I had a look at that fork, and did this: #233

If you'd like to check it over, any feedback would be welcome  :)

realnc:
@jdgleaver Awesome! I can close this PR then.

PR #225, opened 2020-01-19
Visual improvements

sonninnos:
Keeping up with the echodore!

- All relevant keyboard key glyphs (cursors, tab, shifts, Amigas, return, backspace)
- Labelless minimal statusbar modes
- Acceleration for D-Pad mouse

Bonus:
+ Added mouse glyph & revised numpad glyph
+ Doubled VKBD font size in hires double line to make the size match
retroarch_2020_01_20_18_45_58_327


rsn8887:
Thanks!

PR #226, opened 2020-01-22
Fix disk_replace_image_index() return value

jdgleaver:
While updating RetroArch's disk control interface code (https://github.com/libretro/RetroArch/pull/10045) I noticed that the `disk_replace_image_index()` function of this core breaks spec by always returning `false` - this will make 'append disk image' operations fail.

This trivial PR just makes `disk_replace_image_index()` return `true` when successful.


rsn8887:
Thanks!

Issue #153, opened 2019-10-15
Savestate oddity | Core Freeze (Certain Games, Certain Platforms ?)


MajorWilly:
I experienced a savestate that wouldn't load back correctly. I was playing James Pond 2, completed 4 levels, made a savestate. When I tried to load the state the next day, all I got was a black screen and about 1 second of sound (from the beginning of the game!) repeating over and over. Has the temp file been corrupted ?

I noticed that VICE now has a different approach to savestates (no temp files) and wonder if this method will be making it's way to PUAE ?

sonninnos:
Temp files or not will not create or remove issues like this.

So far adf-based saves have worked as expected over here, but with hdf I've noticed some random irregularities, like mouse input skipping massively.

But whatever it is, it won't get fixed without changing/updating the UAE source code.

rsn8887:
AFAIK save states with hdf have never been supported in any UAE, not even WinUAE. They seem to work most of the time but there's no guarantee.

MajorWilly:
I forgot to mention, it was the ADF version I was using. Oh well, I guess it's hit'n'miss.

rsn8887:
Hmm odd, because adf savestates should work.

sonninnos:
Just had a random oddity with the ADF version of Cannon Fodder. Savestate was made at the menu screen were normal saves are made. All was fine to the point where the actual gameplay started, and after that the graphics were all corrupted and messed up (should've taken a screenshot for science). Loaded the same state again and no problems after that.

MajorWilly:
Does that mean it's not loading (or saving) the memory back correctly on a regular basis ?

MajorWilly:
There's definitely something weird. I tried on the PC. I haven't set up Retroarch properly since I re-installed Windows so this is a clean install of PUAE. For a quick test I loaded the ADF version of James Pond 2 (512+512 memory). Did a save state at the beginning of the game, moved him a bit then reloaded the state. I kept getting these flickering green things at the bottom of the sprite -

![James Pond 2 - Codename RoboCod-191016-105755](https://user-images.githubusercontent.com/52862903/66909414-a4fb9f80-f004-11e9-90fc-fb85b70a62fc.png)

Should I leave this issue open ?

sonninnos:
This happening with the same standalone/other PUAE version would be the most interesting bit.

sonninnos:
Also: Does it happen with IPF or a different cracked ADF? And on every "CPU compatibility" setting?

[...]

PR #233, opened 2020-01-29
Remove libco

jdgleaver:
This PR removes the core's dependency on libco. It is mostly derived from hiddenasbestos' puae fork (https://github.com/hiddenasbestos/libretro-uae/tree/gridc), although the implementation there was somewhat incomplete and required a little tweaking  :)

It is difficult to benchmark this core, but from limited testing on my dev machine (using fast forward and monitoring the framerate) this PR seems to improve performance by 8-9%.

Perhaps more importantly, libco is a 'last resort' hack with a number of downsides (as explained by @realnc in PR #224) - removing it is in general a good thing.

@sonninnos I would really appreciate it if you could review this PR. The changes (and the parts I copied from hiddenasbestos) seem rational to me, and everything seems to work correctly - but I don't think this should be merged without the approval of our resident puae expert!


realnc:
Thanks for this!

I gave it a quick test in a bunch of games. No issues so far.

WindyWinston:
Improves performance by 8-9% !! That's got to be a win win.

sonninnos:
Excellent! Looking great, but the CD32 boot animation (with and without disc inserted) ends up in immediate access violation. Loading a savestate before that seems to work fine though, as in CD audio plays and data loads.

I'll soon start a more thorough search. Other major things seem to be as they were (hardfiles work, IPFs work, LHAs work, input reacts on the next frame).

Too bad the performance increase does not seem to help with Soccer Kid AGA, which for my machine still requires dropping frame delay. Normally 12 is fine, but this one needs 10 to run smoothly. It is the only game I've found so far that needs more juice.

jdgleaver:
@sonninnos Awesome, thank you so much for looking at this!

> the CD32 boot animation (with and without disc inserted) ends up in immediate access violation. Loading a savestate before that seems to work fine though, as in CD audio plays and data loads.
> I'll soon start a more thorough search.

That's a strange one! I tested CD32 content quite a bit (that's my main use for the core), and it worked fine on my system - but I'm using Linux, so perhaps that's a factor...? If I can do anything at all to help with your investigations, please let me know (I'll try some ASAN tests when I'm back in the office tomorrow - perhaps there's something I'm missing).

> Too bad the performance increase does not seem to help with Soccer Kid AGA, which for my machine still requires dropping frame delay.

Ah, that's a shame. It may be the case that the performance increase *decreases* with CPU speed - i.e. if you have a beefy CPU, the libco overheads were probably less significant to begin with (if it's any consolation, my hardware is such junk that I have to leave frame delay on zero - so you're already ahead of me  :)  )

I haven't tested on Android yet, but I'm hoping the performance improvement might allow a few more games to run in cycle exact mode (there are some titles that are really borderline at the moment on my tablet and phone),

sonninnos:
My pleasure breh!

I'm pretty baffled with this and have no idea why this occurs only (so far) at the CD32 startup:
```
(gdb) bt full
#0  0x0000000012a99723 in put_sound_word_left (w=0) at sources/src/audio.c:501
        rnew = <optimized out>
        lnew = <optimized out>
        tmp = 0
        rold = <optimized out>
        lold = <optimized out>
#1  sample16si_anti_handler () at sources/src/audio.c:1030
        datas = {0, 0, 0, 0}
        data1 = 0
        data2 = 0
#2  0x0000000012a9d4c1 in update_audio () at sources/src/audio.c:2084
        best_evtime = 41005
        rounded = 41005
        i = 0
        n_cycles = 75219
#3  0x0000000012a9d601 in audio_hsync () at sources/src/audio.c:2131
No locals.
#4  0x0000000012a28e16 in hsync_handler_pre (onvsync=false) at sources/src/custom.c:6432
        hpos = 227
#5  0x0000000012a2a348 in hsync_handler () at sources/src/custom.c:6783
        vs = false
#6  0x0000000012b10004 in do_cycles_slow (cycles_to_add=0) at sources/src/events.c:80
        i = 3
#7  0x0000000012a35c44 in do_cycles_ce (cycles=3072) at sources/src/custom.c:8432
        hpos = 227
        c = 3200
#8  0x0000000012a39406 in cia_wait_pre () at sources/src/cia.c:1513
        div = 512
        cycles = 3072
#9  0x0000000012a3954a in cia_bget (addr=12574464) at sources/src/cia.c:1555
        r = 15
        v = 255 'ÿ'
#10 0x00000000129d7a19 in get_byte (addr=12574464) at ./sources/src/include/memory_uae.h:217
No locals.
#11 mem_access_delay_byte_read_ce020 (addr=12574464) at ./sources/src/include/cpu_prefetch.h:176
        v = 0
#12 get_byte_ce020 (addr=12574464) at ./sources/src/include/cpu_prefetch.h:261
No locals.
#13 0x0000000012f1ac6c in op_0228_21_ff (opcode=558) at sources/src/cpuemu_21.c:1433
        dst = 0 '\000'
        dsta = 12574464
        src = 127 '\177'
        dstreg = 6
#14 0x00000000129ea0cf in docodece020 (opcode=558) at sources/src/newcpu.c:4569
No locals.
#15 m68k_run_2ce () at sources/src/newcpu.c:4666
        r = 0x15935e80 <regs>
        opcode = 558
        exit = false
#16 0x00000000129eeb87 in m68k_go (may_quit=1, resume=1) at sources/src/newcpu.c:5039
        restored = 0
        run_func = 0x129e9b44 <m68k_run_2ce>
#17 0x000000001338352c in retro_run () at libretro/libretro.c:3982
```

Took a while to get any useful readings, because some geezer had put `-s` in the Windows section of the makefile :F

jdgleaver:
@sonninnos Many thanks for the info - but this is a real puzzler...

I've been banging my head against this for hours. I absolutely cannot reproduce the issue.

I tried running the code with AddressSanitizer enabled. This showed up quite a few little errors (I've fixed all of these apart from 3 'trap'-related memory leaks - I don't know the code well enough to handle these) but there was nothing associated with the audio buffer.

It's possible that a couple of these errors *might* have caused subtle memory corruption, so you could try again with the commit I've just pushed - but I doubt it will have an effect...

Now the only way I could generate a crash like the one you see is to hack `audio.c` such that ` mixed_on` is 'false' in `put_sound_word_right()` and 'true' in `put_sound_word_left()`. This offsets the `sndbufpt` pointer and you get out-of-bounds access, just like your backtrace - *but this weird broken condition cannot ever happen*, so it makes no sense.

On the off-chance that something odd is happening on the first frame under Windows, this diff adds a dirty hack that prevents out of bounds audio buffer access:

```diff
diff --git a/sources/src/sounddep/sound.h b/sources/src/sounddep/sound.h
index c9b4f15..11959e8 100644
--- a/sources/src/sounddep/sound.h
+++ b/sources/src/sounddep/sound.h
@@ -49,8 +49,11 @@ void restart_sound_buffer (void);
 
 #define AUDIO_NAME "retroaudio"
 
-#define PUT_SOUND_BYTE(b) do { *(uae_u8 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 1); } while (0)
-#define PUT_SOUND_WORD(b) do { *(uae_u16 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 2); } while (0)
+#define PUT_SOUND_BYTE(b) do { if (((uae_u8 *)sndbufpt - (uae_u8 *)sndbuffer) > (sndbufsize - 1)) {break;} *(uae_u8 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 1); } while (0)
+#define PUT_SOUND_WORD(b) do { if (((uae_u8 *)sndbufpt - (uae_u8 *)sndbuffer) > (sndbufsize - 2)) {break;} *(uae_u16 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 2); } while (0)
+
+//#define PUT_SOUND_BYTE(b) do { *(uae_u8 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 1); } while (0)
+//#define PUT_SOUND_WORD(b) do { *(uae_u16 *)sndbufpt = b; sndbufpt = (uae_u16 *)(((uae_u8 *)sndbufpt) + 2); } while (0)
 //#define PUT_SOUND_WORD(b) PUT_SOUND_BYTE(b << 8); PUT_SOUND_BYTE(b >> 8)
 
 #define PUT_SOUND_WORD_LEFT(b) PUT_SOUND_WORD(b)
```

Maybe you could try this...? I'm honestly completely stuck here, but maybe it will narrow things down a little...?

rsn8887:
Booting into CD32 using sound: exact works on osx. I see the animation. I will test Switch later.

sonninnos:
Sigh, no dice with the new changes. I even tried to put Sound Separation to the default 100% and the same thing happens.

rsn8887:
Tested this PR on Switch starting CD32, works.

jdgleaver:
@rsn8887 Great, thanks for testing! So we know it works on Linux, Mac and Switch (and I guess all unix-like systems). It's only Windows that's messing about.

@sonninnos From your reports so far, it seems to me that the sound buffer pointer is being set or incremented incorrectly on the first frame. So strange! I think some old-fashioned printf() debugging is in order, if you're up for it. We can continue this on Discord.

jdgleaver:
After much debugging (many thanks, @sonninnos!), we discovered a nasty memory corruption error - this was effectively hidden on Linux (due to different memory layout), but caused a crash on Windows because it happened to overwrite the audio buffer pointer.

I've just pushed a commit that should fix this.

If @sonninnos gives it the OK, then the PR should be ready to merge  :)

WindyWinston:
Memory corruption ? I wonder if it will also fix https://github.com/libretro/libretro-uae/issues/153 on Android ? He mentioned it may have been the status bar freezing the core after 35 minutes in certain games. I have also experienced this.

sources/src/include/commpipe.h
```diff
@@ -48,6 +48,11 @@ STATIC_INLINE void destroy_comm_pipe (smp_comm_pipe *p)
     uae_sem_destroy (&p->lock);
     uae_sem_destroy (&p->reader_wait);
     uae_sem_destroy (&p->writer_wait);
+    if (p->data)
+    {
+		free(p->data);
+		p->data = NULL;
+    }
```
realnc: It's not necessary to check for null. `free(NULL)` is always a no-op. This can just be:

    free(p->data);
    p->data = NULL;

libretro/libretro.c
```diff
@@ -2583,26 +2571,38 @@ void retro_init(void)
...
    }
 
    const char *save_dir = NULL;
    if (environ_cb(RETRO_ENVIRONMENT_GET_SAVE_DIRECTORY, &save_dir) && save_dir)
    {
       // If save directory is defined use it, otherwise use system directory
-      retro_save_directory = *save_dir ? save_dir : retro_system_directory;
+      strlcpy(
+            retro_save_directory,
+            !string_is_empty(save_dir) ? save_dir : retro_system_directory,
```
realnc: This form of ternary op check can look rather confusing. It's clearer to write:

    string_is_empty(save_dir) ? retro_system_directory : save_dir

libretro/libretro.c
```diff
@@ -3918,18 +3965,24 @@ void retro_run(void)
       {
          thisframe_y_adjust_update_frame_timer--;
          if (thisframe_y_adjust_update_frame_timer == 0)
+         {
             if (opt_vertical_offset != 0)
+            {
                thisframe_y_adjust = minfirstline + opt_vertical_offset;
+               request_reset_drawing = true;
+            }
+         }
```
realnc: This can just be:

```c
if (thisframe_y_adjust_update_frame_timer == 0 && opt_vertical_offset != 0)
{
   thisframe_y_adjust = minfirstline + opt_vertical_offset;
   request_reset_drawing = true;
}
```

sources/src/cfgfile.c
```diff
@@ -3066,7 +3066,8 @@ static int cfgfile_parse_newfilesys (struct uae_prefs *p, int nr, int type, TCHA
 #ifdef FILESYS
 	add_filesys_config (p, nr, &uci);
 #endif
-//	xfree (str);
+	if (str)
```
realnc: No need to check for null before freeing.

realnc:
No issues for me so far on Linux. The CD32 boot logo is also working.

Great job getting this to work!

sonninnos:
Oh yes, sweet flying A1200 power LED shaded green colors from Windows land. Excelsior!

jdgleaver:
@WindyWinston Yes, it is a possibility that memory corruption caused the freezing on Android (I've also experienced this), but it might also have been due to using libco in the first place (not sure how 'good' the arm implementation of libco was). I'll try to test this myself once the PR is merged and the buildbot spits out an update :)

@realnc My pleasure! Thanks for testing, and for the comments! I'll address these fixes tomorrow.

@sonninnos Awesome - you da man! Your testing is always appreciated  :)

sources/src/include/commpipe.h
jdgleaver: Ah, I come from a time when you did have to check that  :)

I do like the clarity of checking for NULL. I know it's a matter of opinion...

libretro/libretro.c
jdgleaver: Good catch - I agree!

libretro/libretro.c
jdgleaver: Another good one! I'll fix that :)

libretro/libretro.c
jdgleaver: Done!

libretro/libretro.c
jdgleaver: Done!

rsn8887:
Is this ready to merge?

jdgleaver:
Yes, this should be all good now

rsn8887:
Thanks! Merging.

jdgleaver:
Nice! Thank you :)

Issue #241, opened 2020-02-11
IPF disk support broken


mors-gotha:
Hi!
In latest PUAE bilds IPF disks don't work (retroarch linux (centos 7) crashes when a IPF disk is loaded).
IPF disks load correctly in 2020-02-09 build, PUAE (2.6.1 7506965).

Thanks for the incredible work you're doing!

sonninnos:
Works fine with Windows, and I don't see any apparent reason why any change after that commit would affect IPFs. Can any Linux guru confirm?

jdgleaver:
Just tested on openSUSE Leap 15.1, using the latest release of the core + `capsimg.so` v5.1 from here: https://github.com/rsn8887/capsimg/releases

Verified good IPF files load and run just fine (no crashes) - is it possible you might have bad files...?

mors-gotha:
I confirm it doesn't work.
I use the same capsimg.so 5.1 and the IPF disks are an official realease of Team17 (i tried Alien Breed SE, overdrive and superfrog).
Retroarch crashes as soon as i load a IPF disk with the lastest build.
With the 2020-02-09 build it works perfectly

sonninnos:
Sorry I should've been clearer that I need stdout/stderr log, since UAE outputs everything there for the time being. Searching for:
```
capsimg.dll opened
CAPS: library version 5.1
CAPS: type:1 date:24.6.2003 5:57:26 rel:615 rev:1
```

mors-gotha:
How do i create a stdout/stderr log?

sonninnos:
Just run RA in a console with --verbose.

mors-gotha:
Here's the new log ;-)
[retroarch.log](https://github.com/libretro/libretro-uae/files/4188817/retroarch.log)

sonninnos:
OK, looks like it has nothing to do with IPF, and instead you should get a crash with any other image file too. It crashes right at the point of displaying the image name in the statusbar.

mors-gotha:
You're right. Retroarch/PUAE crashes even with ADF disks but it works with lha games.

sonninnos:
Fix will be coming as soon as I get the next PR ready. I'll try to be quick and postpone a few things.

mors-gotha:
Thank you again for your hard work. 
PUAE is one of the best Amiga emulator i've had the luck to try!

sonninnos:
Please confirm if it worked, when possible.

mors-gotha:
[...]

PR #242, opened 2020-02-12
Filesystem fixes

sonninnos:
- Archive & directory access up to par
  = Some non-working LHAs can be taken off the list, such as:
  - BeneathASteelSky_v2.1_CD32
  - DeluxeStripPoker2&DataDisks_v1.0

Should also deal with #241


rsn8887:
Thanks!

PR #244, opened 2020-02-12
Dynamic resolution switching, Analog joystick, WHDLoad.hdf update

### LINKED ISSUE #243: RFE: autobooting with a workbench for the non-bootable hdf images?
It's a bit annoying that there are still hd images around that are not WHDLoad ready, so a option to fallback to loading a workbench if the hd is not bootable and there is no whdload boot file present to manually navigate the workbench to the game and boot it up would be cool.

Not sure if a good idea, since ofc, there are different WB for different models and how savestates would react.

Or we can wait until people stop creating lha/hdf that are not bootable and don't have the whdload file; though the problem appears to be that the people who made these (often hacks/remakes, like the Heart of the Alien Redux) do it and then no-one ever dares to distribute a modified version, even for convenience.

sonninnos:
WHDLoad.hdf:
- Added mathieeedoubbas.library
- Improved scripts
Closes #243 

Other:
- Resolution can be changed without forced reset
- Added analog joystick support
- Removed parenthesis from PAL/NTSC tags for less manual renaming yet again


rsn8887:
Thanks!

PR #245, opened 2020-02-17
Visual upgrades & CPU adjustments

### LINKED ISSUE #235: [Request] Show selected model in status bar
This may be pointless but I notice that there is a lot of blank space in the status bar and wondered if it would be good to show the currently selected model. So when running a game you don't have to check the core options to see what model was set or if it's on automatic you can see what model it chose. And some ECS games have AGA enhancements. (Maybe this could also be done for vice). You could use the same naming as in the readme so you would also know how much memory is set (A500OG, A500+, CD32FR etc.) Just an idea.

Slightly off topic, in the PUAE status bar there are 4 joysticks even though I only have 1 user set in Retroarch ?

sonninnos:
- Added more useful information (resolution, memory, model) to the statusbar
  Closes #235 

- Fixed double line deinterlacing

So instead of this:
retroarch_2020_02_16_23_02_12_755

We get this:
retroarch_2020_02_16_23_01_57_594

And instead of this unreadable mess:
retroarch_2020_02_16_16_01_53_370

We get this:
retroarch_2020_02_16_16_01_59_183

(Yes the before-screenshots are Single Line, but currently Double Line does the same bob deinterlacing, so didn't bother to remove the fix and compile again just for the screenshot)

And of course the complimentary WB shot:
retroarch_2020_02_16_13_02_38_297

Same as a raw RA screenshot with aspect ratio correction:
A1200-200216-130449-aspect

- Separated Resolution and Line Mode core options

- New default Video Resolution: Automatic
  - Defaults to Hires and switches to SuperHires when resolution is sensed
    - Only WB tells the resolution cleanly, so had to resort to hackery to get Super Skidmarks to register
- New default Video Line Mode: Automatic
  - Defaults to Single and switches to Double when interlacing is sensed

Bonus:
- Adjusted 68020 and higher CPU speeds to more match Sysinfo on all CPU compatibility modes
  - Fixes some games with fluctuating screen refresh, like Brutal Sports Football & Flood
  - Surprisingly gives a performance boost for doing less unnecessary work (No need to drop frame delay in Soccer Kid AGA anymore)
  - Please test and search for any negative side effects, as I haven't found one yet!

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ These parameters control the output resolution of the core:
 
 |Name|Values|Default|
 |---|---|---|
-|Video Resolution|Low, High (SL), High (DL), Super-High (SL), Super-High (DL)|High (DL)|
+|Video Resolution|Low, High (SL), High (DL), Super-High (SL), Super-High (DL), Automatic (SL), Automatic (DL)|Automatic (DL)|
 |Video Standard|PAL, NTSC|PAL|
 |Aspect Ratio|PAL, NTSC, Automatic|Automatic|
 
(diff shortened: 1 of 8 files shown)


WindyWinston:
👍 👍 👍 👍 👍

jdgleaver:
This man deserves a knighthood.

Outstanding work (as always!).

sonninnos:
Please no merging yet, I found a small defect of The Settlers menu doing constant Hires-SuperHires-switching..

sonninnos:
OK, removed the rather pointless Super Stardust SuperHires-triggering hack that caused a false positive.

rsn8887:
Ready to merge?

sonninnos:
Aye!

WindyWinston:
Interlace looks awesome ! One thing that I found, the game Pinball Illusions (I used WHDLoad CD32 version for easy controls) has a HiRes mode by pressing Right on the D-Pad or getting a multi ball. Because of the interlace I think you said it halves the framerate so looks a bit choppy. It also still looks a bit interlaced !? Maybe its Super HiRes. Not a real problem though as I prefer the LowRes mode. I don't know if there are any other games that use interlace mode for gameplay, not counting slow gameplay like Darkseed, Sim games etc. where you wouldn't notice.

sonninnos:
Yep, I should rephrase the readme-stuff, since in fact double line laced stuff stay laced and should be deinterlaced afterwards. But since presumably there are no shaders that can double the framerate, looks like I have to copy some more recent stuff from WinUAE, which has a "Double, fields" option alongside this currently only option "Double, frames". That mode outputs stuff that should look better with a deinterlacing shader.

PR #249, opened 2020-02-26
Use memory-based save states

jdgleaver:
At present, the core implements save states using a temporary state file generated by the core itself. This is quite slow, but more importantly it causes a huge amount of unnecessary disk writes - when saving/loading the state of a WHDLoad install, for example, it is very easy to rack up many hundreds of megabytes of written data (not good for flash storage!). Writing to disk like this also precludes frontend features like rewind, readahead, etc.

This PR implements purely memory-based save states, so no more unnecessary disk writes.

**Note 1:** Rewind is quite useable now, provided granularity is set moderately high - 5 frames or so. *However:* rewind does not work reliably/at all for WHDLoad installs (and it can also fail when using an `(MD)` floppy disk playlist)

**Note 2:** Runahead does not work at present. I managed to get it sort-of working for non-`(MD)` floppy games using a couple of hacks, but it was far, far too slow to be useable. More importantly, CD32 games and most WHDLoad installs hang on load content (or segfault) with runahead enabled, and I couldn't determine why. So there's still more to do here...

### DIFF
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -13,6 +13,8 @@
 #include <windows.h>
 #endif
 
+#include "zfile.h"
+
 extern int retrow;
 extern int retroh;
 extern int pix_bytes;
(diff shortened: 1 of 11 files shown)


rsn8887:
Looks good to me, thanks!

PR #251, opened 2020-02-27
WHDLoad update

sonninnos:
- WHDLoad.hdf used as raw files by default
  - Should give less problems with savestates
- Combined WHDLoad.hdf & WHDSaves.hdf core options to "WHDLoad Support"
  - Old "Enabled" is split into "Files" & "HDFs"
- Directories can be launched like LHA/HDF

+ Restored 68020 normal speed to default
+ Fixed keylist goof

As gzip won't handle multiple files, I decided to take the zip route, since it can also be extracted with the existing methods, mostly. Please confirm on all platforms that the WHDLoad-directory appears and populates properly in saves!


jdgleaver:
I can confirm that this works great on Linux and Android.

Very nice addition!

(As well as improving WHD save states, the `Files` method means we no longer have to rewrite the entire 2MB `WHDLoad.hdf` each time we load content - this is nice for saving wear and tear on flash storage)

sonninnos:
Whoops, forgot to add Line Mode core option visibility back when..

rsn8887:
Ready to merge?

sonninnos:
Yes sir!

PR #252, opened 2020-02-27
Save state fixes

jdgleaver:
This is a follow up to PR #249. It includes the following fixes:

- At present, any call to `retro_update_av_info()` that occurs outside of `retro_run()` is ignored (since it is unsafe to invoke this anywhere else). Such calls may happen on the first frame, and when loading a save state - and it turns out that this can cause problems in certain edge cases. For example, when running some NTSC content, we end up with the wrong video dimensions (until the next resolution change). This sort of error was hard to detect! Many apologies for missing it. This PR fixes the issue by deferring any av info updates that occur outside the main runloop to the next call of `retro_run()`.

- It turns out that rewind functionality can break in certain games due to an increase in save state size during runtime. This PR now sets a fixed save state size (determined at load content), so the frontend no longer has to deal with changing buffer sizes. Rewind should now always work in all cases.

**NOTE:** With the fixed size save states, and after some further investigation, I can confirm that *runahead now works in most cases* - **but** only if it is enabled *after* content is fully loaded.

Basically, there is window of ~300 frames at the start of the emulation runloop where we cannot save/load states on adjacent frames. This seems fundamental to the underlying puae code, and I have no idea how to fix it (or even if it can be fixed). But if runahead is enabled after this, it is actually quite useable - although it requires a beefy CPU.

### DIFF
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -20,8 +20,11 @@ extern int retroh;
 extern int pix_bytes;
 extern int zoomed_height;
 extern int imagename_timer;
-extern bool retro_update_av_info(bool, bool);
 extern void reset_drawing();
+
+extern bool retro_request_av_info_update;
+extern bool retro_av_info_change_timing;
+extern bool retro_av_info_is_ntsc;
 
 extern int vkey_pressed;
 extern int vkey_sticky;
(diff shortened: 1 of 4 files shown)


rsn8887:
Nice, thanks. Somehow Travis is pending still?!?

rsn8887:
Travis succeeded although it kept saying “pending” here.

PR #253, opened 2020-02-29
Extended ZIP support

sonninnos:
Blocked the frontend extraction and added some simple logic to determine the content:
- M3U playlist will be created and launched when floppy disks are found
- Hard drive and CD images take precedence and the first one is picked when found
  - Perhaps later combine floppies and hard drive images to work together..?
- If no images are found, the content will be seen as a directory, thus also making zipped WHDLoad installs usable

Please point out if I missed any use case! Any potential problems with the way of clearing out the temp files at deinit?


rsn8887:
Looks good to me, thanks! Merging.

PR #260, opened 2020-03-17
Minor save state improvements

jdgleaver:
This PR makes two small improvements to save state handling:

- When loading floppy disk content via `m3u` files, restoring a save state now updates the frontend with the correct 'currently selected' disk index (previously this was left at whatever index was selected before restoring the state, which was often wrong and made for a confusing user interface...)

- When creating save states, the memory buffer is now created with the correct 'final' size (instead of relying on dynamic allocation, expanding as required). I didn't actually expect this to have much effect, but it turns out all those `realloc`s were quite expensive: previously I could only run at fullspeed with rewind enabled with a granularity of 4-5 frames - now I can drop this down to 3.

### DIFF
--- a/libretro/libretro-glue.h
+++ b/libretro/libretro-glue.h
@@ -14,6 +14,7 @@
 #endif
 
 #include "zfile.h"
+#include "retro_disk_control.h"
 
 extern int retrow;
 extern int retroh;
@@ -39,6 +40,7 @@ extern int retro_max_diwstop;
 extern char retro_system_directory[512];
 
 extern struct zfile *retro_deserialize_file;
+extern dc_storage *retro_dc;
 
 #ifndef _WIN32
 #define TCHAR char /* from sysdeps.h */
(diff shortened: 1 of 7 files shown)


i30817:
This would be very useful in other cores too, are you interested in doing it for cores like beetle?

jdgleaver:
> edit: ok, i probably didn't read this right, it's not 'autochanging the actual index', just updating the ui to the 'current' one.
>
> Dreams of seamless changing disks and showing savestates for the whole game on games with multiple discs : ruined. 

Umm... PUAE already handles this correctly. If you save a state on disk 3, loading that state will put disk 3 in the drive. All that was missing was the fact the frontend would show the wrong index (which made things confusing if you subsequently went to change disks again afterwards)

> edit2: are you planning read-ahead, since you're doing rewind work faster?

Run ahead already works, but with a big caveat - PUAE itself is unable restore a save state within the first ~300 frames or so of booting. This means that if you load content with run ahead enabled, it locks the emulator in a boot loop.

If you load content, wait a few seconds and *then* enable runahead, it works in most cases (although WHDLoad games can be temperamental...), and 'second instance' run ahead is generally fast enough to be useable.

rsn8887:
Awesome, thanks!

PR #261, opened 2020-03-18
VKBD touch control

sonninnos:
Only tested with Windows and mouse, since `RETRO_DEVICE_POINTER` also reacts with it. Hence also disabled real mouse control while VKBD is visible.

Bonus:
- Switched O2 to O3 for a small performance boost. No ill-effects witnessed yet, but standing by for surprises from other platforms.


rsn8887:
Tested on Switch also, works great!!!! merging.

PR #262, opened 2020-03-20
WHDLoad changes (button overrides)

### LINKED ISSUE #258: Could you make a change to the 'whdload.prefs options' splash setting?
'splash delay' -1 to turn on the 'start' button when 'splash' is selected and disable the timeout (i assume 'config' will use the config file in /system as it is, and splash sets some defaults.

https://fs-uae.net/devel/docs/options/whdload-splash-delay.html

The thing is, 'splash' is already indicating you want to select a option, and the current time given is kind of unfriendly to select.

I also read you were supposed to be able to activate this menu on boot (like you can activate the readmes) but i wasn't able to, pressing enter and with a cd32 emulated pad (on a keyboard), which is the whole reason i changed to 'splash'.

Using the setting 'config' and editing the prefs file to set SplashDelay=-1 worked like i intended but i'd prefer that

1. the normal autoboot had a key that worked to start the config menu (not sure this is user error on my end or a conflict of 'amiga keyboard/native keyboard emulating a cd32 controller).
2. in splash mode or the 'autoboot interrupt' mode the menu appearing with that setting to -1 so people could chose the options leisurely not in 2 seconds and it's gone (couldn't do it in even 10 times).

sonninnos:
- New WHDLoad boot overrides (keyboard input removed):
  - Red for Selector
  - Blue for Config
  - Red+Blue for Readme+MkCustom
  - LMB for Splash
  - RMB for Config+Splash
- Renamed "WHDLoad.prefs Options" to "WHDLoad Splash Screen Options" + clarified options
- Updated Readme
Closes #258 

Bonus:
+ 80MB option for HDF creation
+ Fixed directory launching breakage because of a `file_exists` check


rsn8887:
Thanks!

PR #266, opened 2020-04-08
Rewrite maccess.h using new retro_endianness

phcoder:


rsn8887:
I see you are replacing simple things like `*a = v` with function calls and removing assembler optimizations. So can you confirm this doesn't have a negative performance impact?

rsn8887:
I just tested with Reshoot-R on Switch, a game which doesn't achieve 50 fps.

With this PR: 44.4 fps
Before this PR: 44.4 fps.

So seems ok.

PR #267, opened 2020-04-08
Controls + Automatic zoom updates

sonninnos:
Controls:
- In analog joystick mode D-Pad did nothing, therefore it now always controls mouse to minimize manual joy/mouse switching
- Core option for reordering joystick ports
  - Aimed at Arcadia and 4-player adapter games

Automatic zoom:
- Less aggressive operation = the change of geometry now requires first and last drawn lines to stabilize for couple of frames
  - Downside is that there is a minor delay in the zoom, but upside is that there are less unwarranted and pointless geometry changes
  - Mainly affects games with full screen slide animations and other transition gimmicks that trigger the calculations, mostly for no proper reason
    - Clear performance improvements at least in: Sink or Swim, Tiny Skweeks, Pinball Fantasies CD32, Soccer Kid AGA
- Finetuned maximum `thisframe_y_adjust` to allow the bottom right part of Agony intro screen to stay in view

Of course the geometry changes do not really affect a beefy GPU running at max clocks, but it certainly does when running at idle clocks with tweaked frame delay.

And please do let me know if you can find some cases which are still bad or even worse!

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -104,6 +104,7 @@ extern bool opt_keyrahkeypad;
 extern bool opt_keyboard_pass_through;
 extern int opt_cd32pad_options;
 extern int opt_retropad_options;
+extern char opt_joyport_order[5];
 int imagename_timer=0;
 static char statusbar_text[RETRO_PATH_MAX]={0};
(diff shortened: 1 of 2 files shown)


rsn8887:
Looks great, thanks! Merging...

PR #268, opened 2020-04-09
Rewrite libretro m68k.h and m68kops.h to use non-libretro variants

phcoder:
They encapsulate per-cpu optimizations. There is no reason to
have a version for libretro


sonninnos:
As that analyze suggests, building is now very messy with those `implicit declaration`s compared to 0 warnings before. But not only that, as Windows won't build at all anymore.

`ld` spits out a ton of `undefined reference`s to `optflag_*`

phcoder:
I believe it's a problem with analysis. Do you have windows compile log?

sonninnos:
All right, green light from Windowsburg!

rsn8887:
Tested on Switch and Vita, too. Works fine, same fps as before.

PR #271, opened 2020-04-10
Fix Android unzip issue, Remove gcc stripping "-s"

### LINKED ISSUE #269: [Android] Zipped floppies & Whdload 'Files' mode not loading content
I just tried to load a single ADF (zipped) and the Whdload file requester keeps appearing ?!?

sonninnos:
Closes #269 hopefully, please confirm!

Also removed stripping, because the space saved is rather minimal, and the benefit of getting readable results straight from `retroarch_debug.exe` is still more valuable.

Bonus:
- Horizontal centering optimization
  - Fixes Test Drive II intro crash when the Porsche is desperately trying to be centered
- Core option for disabling floppy sounds with empty drives

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,6 @@ ifeq ($(DEBUG), 1)
    CFLAGS += -O0 -g
 else
    CFLAGS += -O3
-   SHARED += -s
 endif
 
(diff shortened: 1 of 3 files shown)


rsn8887:
Thanks, looks good!

PR #273, opened 2020-04-11
Android unzippings yet again

sonninnos:
The last Android fix was half baked and had other problems as well. Brainstorm credits to @jdgleaver


rsn8887:
Thanks!

Issue #272, opened 2020-04-11
Xenon 2 CDTV failure to load, except when model is CD32


KMFDManic:
Xenon 2 CDTV Version seems to get stuck on Black Screen with a little red dot displaying at top left.  Most other games loaded absolutely fine for me!  And, Space Wars did not seem to allow me any controls, aside from the Xenon 2 Test.  Curious if you guys get the same results on your end on these 2 specific games.  Xenon 2 is quite possibly the best game on CDTV, bar none.  So, would be cool if there is a conflict, we can figure out what causes it!

sonninnos:
I just recently tried Space Wars and it worked fine. Gameplay is pretty bad though compared to Star Control.

That red dot seems to be the mouse cursor, and any keys do nothing. No idea what causes it to get stuck there, but it does not happen in CD32 mode.

KMFDManic:
@sonninnos thanks for the quick reply. I will try Space Wars again.  I've opted for Amiga Xenon 2.  But, it would be beautiful if we could sort the CDTV Version out so we can enjoy the amazing soundtrack!  

Prehistorik and Town With No Name also worked fantastic.

sonninnos:
I'll see what I can do, but since the same version works fine with CD32, it is not high on the list.

KMFDManic:
@sonninnos if forcing it to CD32 works, then definitely don't make it a priority.  I was just curious about what conflict might prevent it from working.  I believe same issue was present in the PC Emulator, too.

Gravity-Grave:
I am also having this same problem with Xenon 2. I used to be able to boot it successfully with the CD32 bios, before the CDTV bios was introduced. But now, when I select to load with either of the two CD32 model options, the system still seems to default back to the black CDTV loading screen. If I try deleting the CDTV kickstart file from the system folder, all CDTV games then refuse to load, giving me a 'Failed to load content' notification. I have also tried toggling through the Cycle-exact, Normal and More Compatible CPU options, but they don’t appear to make a difference.
I can’t find a way to force CD32 anymore. Is there something I am doing wrong?
I am on a Windows PC, Retroarch 1.8.5, PUAE (2.6.1 bd2b98f)

[...]

PR #274, opened 2020-04-11
Change model force hierarchy, Option label updates

sonninnos:
Currently filename tags are having the last word when the model is chosen, and that conflicts with the core option somewhat.

For example when trying to force a CDTV game to CD32 mode via core options if the file is in a directory that matches the tag "CDTV", it can't be done because the filename matching actually searches the whole path.

This PR changes the pecking order to model core option having the final say, so that game-based overrides can be used whatever the case may be. Filename tags are searched only with the "Automatic" model, like Guru intended.

Any possible blindspots? Seems to working fine. Relates to #272 but does not close.

Also some minimal bonus rewordings and rephrasings to Readme, verbose log and option labels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,11 @@ Single Line combines interlaced fields into one field, which will keep the full 
 - Automatic Line Mode defaults to Single Line and selects Double Line on interlaced screens.
 
 ## Games that require specific hardware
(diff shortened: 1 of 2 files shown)


rsn8887:
Looks good to me! Merging.

PR #275, opened 2020-04-13
CD startup delayed insert, Autozoom improvement

### LINKED ISSUE #272: Xenon 2 CDTV failure to load, except when model is CD32

sonninnos:
- Added a core option for delayed CD insert on startup for those pesky games that won't run when powering on with disc inserted, even on the real thing (Xenon 2 CDTV, Total Carnage CD32)
  Closes #272 
- Yet another pointless work reducing autozoom improvement:
  - Skip geometry update if the used values return back to the original values during the stabilization period (Lethal Weapon screen fades)
- Fixed sound filter type not updating on startup, which seems to have been broken for a long time, from libco removal perhaps

Bonus:
- Found a way to remove my hacks to get CD eject/insert detection working by adding `#define SCSIEMU`, because only then was `blkdev_vsync()` run, which does the necessary `check_changes()` automatically

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -36,7 +36,6 @@
 
 int libretro_runloop_active = 0;
 
-extern void check_changes(int unitnum);
 extern int frame_redraw_necessary;
 extern int bplcon0;
 extern int diwlastword_total;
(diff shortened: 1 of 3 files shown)


rsn8887:
Thanks!

PR #276, opened 2020-04-15
Backport double line field interlace

sonninnos:
New stuff:
- Replaced current framerate halving double line frame with full framerate double line field
- Added core option "Remove Interlace Artifacts" which behaves best on static screens
  - Confusingly the option is named `gfx_flickerfixer` in confs and `gfx_scandoubler` internally, but according to the changelog it is neither:
> - "Flicker fixer" renamed to "Remove interlace artifacts" (it really
  isn't same as flickerfixer or scandoubler)

Fixes:
- Sound filter type option was accidentally forced at "Automatic"
- More autozoom+centering solutions
  - Allow UAE internal centerings but prevent RA geometry sets in certain situations
- Random crash when changing resolution and interlacing at the same time in WB

Bonus:
- Core option unifications and reorganizings


rsn8887:
Looks good, thanks! Merging.

PR #277, opened 2020-04-15
Fix calling init_custom() for no reason

sonninnos:
Just a small fix this time. The right answer was right in front of me but I took a wrong turn which didn't even work correctly, sorry about that.


rsn8887:
Thanks! Merging.

PR #279, opened 2020-04-22
M3U ZIP fix

### LINKED ISSUE #278: Zipped disks in M3U don't mount
If you load a M3U and the ADFs are zipped they don't seem to mount.

sonninnos:
Closes #278 

Also bonus fixed Browse Archive on Load Content.


rsn8887:
THanks!!!

Issue #215, opened 2020-01-12
Need to restart restroarch after using lr-puae after closing content (switch)


tomjramsden:
Using retroarch 1.8.2 and PUAE 2.6.1 7abc7f1

However seen this behaviour for more than a few releases:
I haven't seen this behaviour with any other core I am using...

Steps to replicate
1. Launch game (typically HDF or lha from playlist or load content)
2. Finish game and 'close content'
3. Launch another game (including other platforms)
4. Retroarch will crash (the software was closed because an error occured)
Alternatively if I try and 'restart retroarch' from the main menu it will also crash

If I restart retroarch and load another game it will work fine.

**I will try and get some verbose logging from retroarch to see if it helps troubleshoot/diagnose.**

rsn8887:
This is a known problem on static builds like Switch. PUAE is not the only core with this problem. Dosbox-SVN has the same issue, for example. The problem happens because, on static builds the core has to cleanly de-init and reset ALL internally used vars when retro-deinit is called. If the core uses any static vars internally all of those would have to be reset to their correct initial values.

tomjramsden:
> AFAIK this is a known problem on static chainloaded builds like Switch. PUAE is not the only core with this problem. Dosbox-SVN has the same issue, for example. The problem happens because, on static builds, the core should cleanly de-init and reset ALL internally used vars when retro-deinit is called. If the core uses any static vars internally all of those have to be reset to their correct initial values to prevent this crash.

Thanks for clarifying. Is there anything on the roadmap to address this in the future?

rsn8887:
If you want to sift through every line of PUAE core code finding all the static vars etc be my guest.

tomjramsden:
> On Vice core there was a workaround. A similar workaround might be possible here. It would involve checking on load content if the main loop of the emulator is already running and then acting accordingly.

Nice one! Sounds like an interesting idea! Hopefully you’ll some traction with it. Been playing around with Vice recently, it’s an amazing core, it just works with everything I throw at it and works so well on the switch.

rsn8887:
Fixed.

PR #280, opened 2020-04-23
Attempt to fix crash when reloading core on static builds

jdgleaver:
At present, the core will crash the second time content is loaded when using a static build (Switch, etc.). I don't have any supported hardware to test this on, but by building under Linux using the `--enable-preserve_dylib` configure option I was able reproduce a segfault and graphical corruption caused by static variables carrying over incorrect values from the previous run.

This trivial PR 'resets' these static variables appropriately. I hope this will fix issue #215 - if anyone could test a static build, I would be grateful!

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -3090,6 +3090,9 @@ void retro_deinit(void)
    // Clean ZIP temp
    if (retro_temp_directory && path_is_directory(retro_temp_directory))
       remove_recurse(retro_temp_directory);
+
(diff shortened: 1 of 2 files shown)


rsn8887:
Tested on Switch - it works! Thank you so much for fixing this. And thanks for finding out how these kinds of problems can be debugged on linux.

jdgleaver:
Thank you for testing! I'm so glad it works :)

PR #281, opened 2020-04-25
UX + behind the curtain maintenance

sonninnos:
Major:
- More convenient error messages for the user when Kickstart ROM or CAPSImg library is not found
- Changed default "CPU Compatibility" to "Normal" on all platforms
  - Please update the list accordingly @WindyWinston
- Enabled floppy sound emulation by default with 20% volume

Minor:
- More accurate initial PAL/NTSC Hz
- Removed "Sound Output" core option, because:
  - "Normal" and "Exact" have no distinction in the code, WinUAE does not have the "Normal" option at all, and the other options are also useless
- Increased sound buffer, because I witnessed no extra delay, and perhaps it helps a bit with low power machines

Bonus:
- Reorganized `libretro-common` to proper form
- Makefile tidy-up


rsn8887:
Thank you! Merging.

PR #284, opened 2020-04-29
Temp M3U sorting, WHDLoad update

sonninnos:
Regular:
- Sorting for generated M3Us, because `readdir` does not guarantee alphabetic order on all filesystems
- WHDLoad helper updated to also copy `kick40063.A600`, which also required increasing the HDF size (both HDFs to 4MB)
- Statusbar updated to match VICE behavior 

Bonus:
- Renamed `libretro.c` to `libretro-core.c` for the obvious reason
- Readme updates, such as lots of obsolete WHDLoad info removed


rsn8887:
Thanks!

PR #287, opened 2020-05-07
Model preset overrides, Optional region search

### LINKED ISSUE #285: Feature Request Idea: Variable Frame Skip PER Model Type
Everything has been absolutely fantastic, bar none.  But, I got to thinking of one thing that would really be a nice addition.  On lower spec platforms, AGA games clearly won't run well unless frame skip is applied.  But, if you do a "universal" frame skip, it will affect games that do not need it.  Maybe, just maybe, if it doesn't complicate the code set all too much, Core Options for Variable Frame Skips Per Model Type, such as AGA = 0, 1, 2, and so on.  With Model Detection already in place automatically or recursively, this would be beautiful indeed to be able to run Aladdin AGA with frame skip and Giana Sisters without, without needing to muck about with anything that can create a flustercluck or sorts.

Thanks as always, both of you @sonninnos and @rsn8887 for taking into account any of my feedback and ideas!  It has truly made the entire Amiga/Commodore 64 thing all that much cooler!  I am still absolutely glad that @fr500 entertained my keyboard to controller mapping idea a few years back!  That is still, by far, one of the absolute best additions ever added for many of the PC Cores that don't have keyboard accessibility!

Thanks again, as always!  No complaints otherwise.  Just trying to think of anything that can be enjoyed by others on EVERY platform:)

And, of course, anything needing tested, I can compile immediately and verify on my end.

sonninnos:
Major:
- Model presets can be customized with config files
  Closes #285 
- Region forcing via filename tags is optional, default is enabled

Minor:
- No-Intro tags added to region forcing
- Corrections to real NTSC behavior (aspect ratio toggle, statusbar position)

Bonus:
- Core option for defaulting/setting D-Pad control mode (Joystick/Mouse)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,8 @@ The whole path (filename and directory) will be searched for the following tags 
 | |**x**|**CDTV**|Amiga CDTV, 1MB Chip RAM|
 | |**x**|**(CD32)** or **(CD32NF)**|Amiga CD32, 2MB Chip RAM|
 | |**x**|**(CD32FR)** or **FastRAM**|Amiga CD32, 2MB Chip RAM + 8MB Fast RAM|
-|**x**|**x**|**NTSC**|NTSC 60Hz|
-|**x**|**x**|**PAL**|PAL 50Hz|
+|**x**|**x**|**NTSC** or **(USA)**|NTSC 60Hz|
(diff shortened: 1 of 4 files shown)


sonninnos:
Please don't merge yet, I found a goof!

sonninnos:
Phew, it was just slave related, all clear!

rsn8887:
Ok! Thanks!

PR #291, opened 2020-05-21
Amiga Forever legal BIOS naming conventions

KMFDManic:
For those who legally purchase BIOS from Amiga Forever, here are the appropriate naming conventions.

With the dozens of BIOS that come in the Amiga Forever Plus Edition (minimal purchasable to acquire the necessary BIOS), it can become a little confusing for some knowing what to rename to what.  Tested and Approved with latest Plus Edition!  

The interesting difference is that older Amiga Forever Releases actually required a rom.key, as a security check, in order to utilize the BIOS.  Removal of the rom.key necessitation in more recent Amiga Forever Releases pretty much identically approximates the md5s for both sets, aside from the missing combination KS + extended v3.1 rev 40.060 AKA kick40060.CD32


rsn8887:
Thanks!

PR #292, opened 2020-05-22
Zoom mode rework, Remapping fixes

sonninnos:
Major:
- Zoom mode can also do horizontal cropping
  - Horizontal + vertical enabled by default
  - Please confirm that VKBD touch is bueno with all sizes!
- Remapping adjustments
  - Special hotkeys can override VKBD functions

Bonus:
- Prefixes for hidden core options added
- Improved M3U generation sorting
- Waiting blits disabled by default
- Markdown oopsies in README fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ The whole path (filename and directory) will be searched for the following tags 
 | |**x**|**(CD32)** or **(CD32NF)**|Amiga CD32, 2MB Chip RAM|
 | |**x**|**(CD32FR)** or **FastRAM**|Amiga CD32, 2MB Chip RAM + 8MB Fast RAM|
 |**x**|**x**|**NTSC** or **(USA)**|NTSC 60Hz|
-|**x**|**x**|**PAL** or **(Europe)** or **(Denmark|Finland|France|Germany|Italy|Spain|Sweden)**|PAL 50Hz|
+|**x**|**x**|**PAL** or **(Europe)** or **(Denmark\|Finland\|France\|Germany\|Italy\|Spain\|Sweden)**|PAL 50Hz|
 
 Example: When launching "Alien Breed 2 AGA.hdf" or "AGA/Alien Breed 2.hdf" the model will be Amiga 1200.
 
--- a/libretro/libretro-core.c
+++ b/libretro/libretro-core.c
@@ -526,7 +526,7 @@ void retro_set_environment(retro_environment_t cb)
             { "waiting", "Wait for Blitter" },
             { NULL, NULL },
          },
-         "waiting"
+         "false"
       },
       {
          "puae_collision_level",
--- a/libretro/libretro-core.c
+++ b/libretro/libretro-core.c
@@ -317,7 +317,7 @@ void retro_set_environment(retro_environment_t cb)
       },
       {
          "puae_video_allow_hz_change",
-         "Allow PAL/NTSC Hz Change",
+         "Video > Allow PAL/NTSC Hz Change",
          "Let Amiga decide the exact output Hz.",
          {
             { "enabled", NULL },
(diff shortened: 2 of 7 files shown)


rsn8887:
Looks good! Thanks!

WindyWinston:
Just out of interest, why is 'Blitter OFF' now the default ? Did you find some incompatibilities ? I notice in WinUAE that CD32 & A1200 has '`Wait for Blitter`' always on and A500/A600 only turn it off when using Cycle Exact.

sonninnos:
The goal was to maximize performance, but maybe it was for nought, since now that I've tried with more cases and with a different measuring method, the waiting gives sometimes marginally better results, nngh.

I'll change it back if it is demonstrably worse now. So please if you may, also keep it off for scouting purposes. The waiting doesn't seem to fix any gfx problems with CD32 that I've found though. And the only case so far that I've found that needs the waiting was the old Rick Dangerous slave running on A1200.

WindyWinston:
Personally, I can't say which is better. I've always had 'Wait for Blitter' on and only use Cycle Exact when it's explicitly needed (for issues listed in the 'Don't work with default settings thread and Android to get maximum speed).

rsn8887:
If someone wants to, they could try some of these games that had flickering graphics due to blitter issues for me in UAE4All2 unless I set Blitter: Normal (which I think means wait):
Blastar,
PP Hammer,
Golden Axe,
Mister and Misses.
Rock'n'Roll,
Shufflepuck Cafe,
Spindizzy Worlds.

sonninnos:
Some quick tests with WHDLoads:

- Blastar flickers with immediate blitter on both A600 and A1200, both show no difference between disabled and automatic waits, A600 does not flicker and A1200 flickers, and with A1200 only "Cycle-exact" cures it.

- PP Hammer flickers with immediate blitter on both A600 and A1200, and both are ok with either disabled and automatic waits.

- Golden Axe is über fast with immediate blitter, and also no difference between disabled and automatic waits with both A600 and A1200. No flickering witnessed though.

Maybe I should try changing that "automatic" to "always" and also perhaps try to expose and separate "blitter_cycle_exact" from the combined "cycle_exact"..

WindyWinston:
> Maybe I should try changing that "automatic" to "always" and also perhaps try to expose and separate "blitter_cycle_exact" from the combined "cycle_exact"..

Just for clarity, when you say "automatic" do you mean 'wait for blitter' ?

What is 'blitter_cycle_exact' ? Is it the same as the 'Cycle Exact/DMA' option in WinUAE ? If so then it would be good if you could make it 2 options like WinUAE has as 'Cycle Exact Full' takes a lot of processing power on shite devices.

sonninnos:
Yes, the current "Wait for Blitter" is `waiting_blits=automatic`.

"Cycle-exact" toggles both `blitter_cycle_exact` and `cpu_cycle_exact`. Since all the flickering stuff is related to blitter, maybe it could serve a purpose to have it enabled when CPU CE is off.

And no, `blitter_cycle_exact` is not the same as `cpu_memory_cycle_exact` (Cycle-exact DMA/Memory accesses). There is no `cpu_memory_cycle_exact`  option in this old UAE. I would have exposed it ages ago if it did, and it has been in the TODO to find out if it could be backported somehow.

WindyWinston:
I wonder if `blitter_cycle_exact` would fix the flickering in 'It came from the desert' without having to use 'Cycle Exact' ? That's one to test if you manage to split the option.

i30817:
I always use whdload and it often solves incompatibilities like this. For it came from the desert it mentions:
 - more CPU-dependent loops removed: game is now slower on fast machines
 - fixed blitter wait

sonninnos:
We **are** talking about WHDLoad installs. ICFTD does not work properly in either A600 or A1200 with the current default settings. Cinemaware games tend to need Cycle-exact.

PR #293, opened 2020-05-30
Autozoom improvements, Content path unicode

sonninnos:
Regular:
- Autozoom improvements
  - The earthquake effect in "Lollypop" no longer resizes the geometry
  - Please let me know at once if something else is or got broken!
- Content path can include non-ascii characters like `å` `ä` `ö`

Bonus:
- "Wait for Blitter" restored as the default
- Logging noise reduced


rsn8887:
Thanks!

PR #295, opened 2020-06-02
Zoom centering fixes, Font updates

sonninnos:
I goofed the CD32 boot screen (disc ejected) with the latest autozoom fix, oops.

Also some desperate cleaning regarding `RPATH` and `uae_restart`, which at least will not create a "././"-prefix with Windows anymore. Most likely won't fix anything, and also hopefully does not break anything.


rsn8887:
Thanks!

PR #301, opened 2020-07-15
Directory filesystem fix, Control maintenance

sonninnos:
Major:
- Updated directory filesystem with illegal filenames
  - In practice fixes the Space Taxi 3 input config screen in Windows directory mode, because `con` needs to be named `__uae___con` on the host and vice versa in the emulation
retroarch_2020_07_15_23_53_23_185
- Allowed launching extracted `.slave` files
  - Does not actually launch the correct slave when there are multiple slaves, so for now it just does the same as launching the directory!

Minor:
- Added RetroPad B to turbo fire options
- Defaulted RetroPad X to Space
- Defaulted RetroPad mapping options visibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ And of course for the RetroArch/Libretro team: "http://www.libretro.com/"
 |---|---|
 |B|Fire button 1 / Red|
 |A|Fire button 2 / Blue|
+|X|Space|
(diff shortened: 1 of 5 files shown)


rsn8887:
I almost missed this. Merging. Thanks!

PR #303, opened 2020-07-24
Zoom preset adjustments, .info launching

sonninnos:
- Widescreen zoom presets adjusted to match VICE core behavior, as in instead of increasing height, it decreases width when things get tight
- Addition to previous `.slave` launching to also accept `.info` (directory and slave)


rsn8887:
Thanks! Merging.

PR #305, opened 2020-08-10
(Oh No! More) Directory filesystem fixes

### LINKED ISSUE #304: ZIP disks not mounting in .uae files
Hi,

Similar to:

https://github.com/libretro/libretro-uae/issues/278

I seem to have this issue when trying to mount ZIP disks (of ADF files) referenced in a .uae file - i.e. it doesn't mount.

This is using Retroarch 1.8.4 with  P-UAE Xtreme (June 21st 2020) core on a SNES Classic Mini.

Is this the same issue as https://github.com/libretro/libretro-uae/issues/278 ?

Thank you.

sonninnos:
- Removed redundant leftover steps (`tmp_nname` shenanigans) during initial feature inclusion, which caused random crashes in some cases
- Fixed heap block past requested size -errors if Amiga is rebooted after accessing a `__uae___`-prefixed file and the same file is accessed again
- Fixed illegal filesystem character handling for directories

Some WHDLoad slaves such as "Pinball Dreams" and "Hook" will use rather harsh naming conventions for save directories when saves are redirected using `SavePath` (mandatory for readonly-LHA launching):
- `-->PinballDreams<--`
- `->HOOK<-`

Which means these need to be renamed and evaded, and the current maneuvers were only targeting the file named `con` in Space Taxi 3.

If WHDLoad installs are used as extracted files or if the core option is in HDF-mode, this will not be a problem, so I never would have noticed these on my own, since I use extracted files for serious gaming.

Credits for noticing the monsters and for teaming up to destroy them goes to @jdgleaver 

Bonus:
- Enabled and fixed using ZIPs via .uae confs (Both `floppyX=` and `filesystem2=`)
  Closes #304

### DIFF
--- a/retrodep/sysconfig.h
+++ b/retrodep/sysconfig.h
@@ -7,7 +7,7 @@
 #define CAPS /* CAPS-image support */
 #define FDI2RAW /* FDI 1.0 and 2.x image support */
 //#define WITH_CHD
-//#define A_ZIP
+#define A_ZIP
 //#define A_RAR
 //#define A_7Z
 #define A_LHA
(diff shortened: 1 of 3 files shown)


rsn8887:
Thank you! Merging...

PR #325, opened 2020-10-19
PS3 build fix and issue fix #283

### LINKED ISSUE #283: Wii U build not working from 10/04 onwards
Hi, I am using nightly from 8th April which works fine but from 10/04 onwards the kickstarts do not load on the Wii U, there is a red screen which changes to black but kickstarts do not load, tested a500, a600, a1200.

crystalct:


sonninnos:
Thanks for this! However I think it would be wise to avoid editing anything under `libretro-common`. At least anything in `libretro.h` AFAIK should be moved elsewhere, and the others does not seem to be that necessary either. Also I assumed `jni` stuff is for Android only, so is there some other way for that too?

Also please squash the commits. I can clean out those lha + caps exceptions later on.

What do you say @rsn8887 ?

libretro-common/include/retro_endianness.h
```diff
@@ -95,6 +95,8 @@
 #  endif
 #elif defined (__i386__) || defined(__x86_64__)
 #  define LSB_FIRST 1
+#elif defined (__CELLOS_LV2__)
```
sonninnos, line 98: I think this can be handled in the makefile?

libretro-common/include/retro_miscellaneous.h
```diff
@@ -77,7 +77,7 @@ static INLINE bool bits_any_set(uint32_t* ptr, uint32_t count)
 #ifndef PATH_MAX_LENGTH
 #if defined(__CELLOS_LV2__)
 #define PATH_MAX_LENGTH CELL_FS_MAX_FS_PATH_LENGTH
-#elif defined(_XBOX1) || defined(_3DS) || defined(PSP) || defined(PS2) || defined(GEKKO)|| defined(WIIU) || defined(ORBIS)
+#elif defined(_XBOX1) || defined(_3DS) || defined(PSP) || defined(PS2) || defined(GEKKO)|| defined(WIIU) || defined(ORBIS) || defined(__CELLOS_LV2__)
```
sonninnos, line 80: Also this define is just a few rows earlier.

libretro-common/vfs/vfs_implementation.c
```diff
@@ -282,7 +282,7 @@ int64_t retro_vfs_file_seek_internal(libretro_vfs_implementation_file *stream, i
 libretro_vfs_implementation_file *retro_vfs_file_open_impl(
       const char *path, unsigned mode, unsigned hints)
 {
-   int                                flags = 0;
+    int                                flags = 0;
```
sonninnos, line 285: Heh, not necessary.

crystalct:
I will follow your directives.

libretro/libretro-mapper.c
```diff
@@ -23,12 +23,13 @@
 #include <time.h>
 #define usleep sys_timer_usleep
 
-void gettimeofday (struct timeval *tv, void *blah)
+int gettimeofday (struct timeval *tv, void *blah)
```
sonninnos, line 26: Why is this required? Not that it matters if it does not break anything, just curious.

sources/src/fsdb_unix.c
```diff
@@ -14,6 +14,10 @@
 #include "fsdb.h"
 #include "misc.h"
 
+#ifdef __CELLOS_LV2__
```
sonninnos, line 17: Does not really matter, but would be cleaner if this was inside __LIBRETRO__ below.

crystalct:
Is it possible make 'ps3-deps' folder? I could put header file and PS3 functions there.

sonninnos:
Sounds like a good idea to me, since there already is one for wiiu.

libretro/libretro-mapper.c
crystalct: if (gettimeofday (&tv, &tz) < 0)

crystalct: libretro/libretro-core.c: In function 'ftime':
libretro/libretro-core.c:5617: error: void value not ignored as it ought to be

libretro-common/include/retro_miscellaneous.h
crystalct: Done

libretro-common/vfs/vfs_implementation.c
crystalct: Done

libretro-common/include/retro_endianness.h
crystalct: Done

sources/src/fsdb_unix.c
crystalct: Done

sonninnos:
All right, looking much better. Now just please squash it into one single commit.

crystalct:
@sonninnos  a question about CAPSIMG libraries and PS3.
PS3 support only static libraries so they must be added during compilation  as sources or as binary lib in linking. 
CAPSIMG are free for non commercial use, so i suppose we can add them here. What is your opinion?

sonninnos:
I'm under the impression that including the library is not legal, and that is why I gave up adding it.. I don't know about the details why that is, sorry, GPL something something, heh.

crystalct:
I read http://www.softpres.org/download and they talk about “SPS Freeware License Agreement”
`If you wish to add IPF image access functionality to a third party application, or perhaps you are just compiling support into an application (UAE for example), this is the package you need. When adding support to an application, it is highly recommended that you use the “Access API” below to dynamically link your application to the library, since the library changes periodically. Commercial applications need to use the Access API below because of licensing restrictions.`
The legal difference is shared library versus static libary?

crystalct:
An option Makefile variant for PS3 to download external sources and compile on the fly capslibs could be an idea.

sonninnos:
I (can't say|don't have strong feelings) one way or the other. You should probably consult higher-ups.

Is there any emulator out there that has the library built-in btw? If even WinUAE does not, I guess there is a reason for it..

crystalct:
I'm even not sure that my point of view is correct.... we were just philosophizing. ^_^
On systems like windows, linux and apple, is really simple, the end user download a file, put it in the right folder and that is enough. So, there is not need to add external lib as built-in that is not 100% freeware. That works because the emulator, or the core, is able to work with that library, it not contain the library. But on systems that works differently like PS3, wiiu, psvita, etc etc, that needs statically linked all libs? In these systems even libretro cores are external libraries that must be linked statically to retroarch layer to make a single binary file. End user cannot do anything. He must be a developer, he must customize various sources and build all.
My point of view is help end user by giving him an easy way, or the simplest way, to make that process, maybe so easy that he could not be an developer or a sysadmin, but just someone capable to download a development environment, setup it and run a make command.
A good HOWTO could be enough, but there really is a difference with an optional parameter (even not as default) in a Makefile that download sources a build a library on the fly?
Talking about capsimg, when i pathed E-UAE 0.8.29 r10 -> r11, adding capsimg library, i sent a mail to softpres.org to inform them about that (look at [http://www.softpres.org/?id=ticket&op=view&tid=06247&code=1Smd7DRNvz](url) ), they never told me something to remove it or a legal warning.

crystalct:
DOh!!!
image

Something is changed, in better: http://kryoflux.com/?page=download
Sources are public and licence is this:
[LICENCE.txt](https://github.com/libretro/libretro-uae/files/5435812/LICENCE.txt)

PR #351, opened 2020-11-30
fix invalid unsigned arithmetic.

ihsinme:
your check is incorrect. since the variables are unsigned, it is equivalent to sizeof (outbuf) != zs.avail_out.
so I suggest a simple fix for the error.

### DIFF
--- a/sources/src/zfile.c
+++ b/sources/src/zfile.c
@@ -2268,7 +2268,7 @@ int zfile_zcompress (struct zfile *f, void *src, int size)
 		zs.next_out = outbuf;
 		zs.avail_out = sizeof (outbuf);
 		v = deflate (&zs, Z_NO_FLUSH | Z_FINISH);
-		if (sizeof (outbuf) - zs.avail_out > 0)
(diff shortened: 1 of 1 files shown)


sonninnos:
What will this fix exactly? That is like that even in latest WinUAE..

It is in `zfile_zcompress`, and we are not doing nor needing any compression activities here anyway, only uncompression.

ihsinme:
the only thing I'm talking about is the unsigned conversion problem. which is valid in your code

https://wiki.sei.cmu.edu/confluence/display/c/INT02-C.+Understand+integer+conversion+rules

```cpp
sizeArray = 60; (sizeArray - 10) = 50; true
sizeArray = 10; (sizeArray - 10) = 0; false
sizeArray = 1; (sizeArray - 10) = 18446744073709551607; true
sizeArray = 0; (sizeArray - 10) = 18446744073709551606; true
if (sizeArray - 10 > 0)
```

sonninnos:
Ok, but where is the problem, as in what will it fix that does not work at the moment? Also that is not "our" code, it is UAE code.

PR #421, opened 2021-05-17
Expose chip RAM

leiradel:
Exposes the chip RAM as system RAM for cheats and achievements.

### DIFF
--- a/libretro/libretro-core.c
+++ b/libretro/libretro-core.c
@@ -21,6 +21,7 @@
 #include "blkdev.h"
 #include "disk.h"
 #include "gui.h"
+#include "memory_uae.h"
 
 unsigned int libretro_runloop_active = 0;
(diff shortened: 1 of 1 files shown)


sonninnos:
Cool. Perhaps that env set should be in `retro_init()` with the other ones, for clarity, since:
> * Can be called from retro_init and retro_load_game.

leiradel:
> Perhaps that env set should be in `retro_init()`

Some cores only allocate the required memory after they're given the actual content, that's why I've put it in `retro_load_game`.

I'm not saying that this is the case with P-UAE, only that it was a safe bet.

sonninnos:
Oh ok, as long as it works then.

leiradel:
> And shouldn't this be added too?

Sorry, forgot to answer that question. I don't think this env call is relevant anymore. @twinaphex ?

PR #456, opened 2021-09-09
Emulated mouse + VKBD color adjustments

sonninnos:
Also added `(512KB)` tag and improved video standard option operation with file tag forcing.


i30817:
The mouse has been smoother on my tests. Thanks for this.

PR #472, opened 2021-11-27
Third "Allow Hz" mode + misc

sonninnos:
Video related:
- "Locked PAL/NTSC" option which forces rate to 50/59.94 regardless of interlace state so that audio buffer stays in line, making interlace toggling seamless. It maybe should be the new default, but no reason to rush yet.
- Geometry adjustments

Other:
- Added joystick directions to core option RetroPad mapper
- Removed possibility to add the same disk to DC list
- Adjusted core option sublabels

### DIFF
--- a/libretro/libretro-core.c
+++ b/libretro/libretro-core.c
@@ -6591,9 +6591,12 @@ static bool retro_update_av_info(void)
       /* If still no change */
       if (video_config_old == video_config && retro_refresh == hz)
       {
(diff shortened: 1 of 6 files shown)


zzedmore:
This is awesome. +1 for making 'forced Pal/NTSC' the new default. I used to wonder why the sound muted sometimes when the resolution/interlace changed. I think this fixes it.

PR #490, opened 2022-01-14
PS3: build fix and .lha load fix

crystalct:


sonninnos:
Hmm, why is that `deps/libchdr/src/libchdr_chd.c` change necessary?

crystalct:
Both `7zTypes.h` and `zlib.h` have `typedef unsigned char Byte;` 
In gcc 4.x that causes a compile error (probably in modern GCCs it's a warining).
But `7zTypes.h` has this:
```
#ifndef ZCONF_H
typedef unsigned char Byte;
#endif
```
So just put `#include "zlib.h"` before  #include #include "LzmaEnc.h" (that contains`#include "7zTypes.h"`) to avoid that error.
Simple.

sonninnos:
Okeydokey, Windows and Linux have been warning-free for a while.

PR #499, opened 2022-03-24
Don't upload audio to frontend multiple times per frame

realnc:
It increases the chance of audio dropouts. Also, it's better to submit
audio to the fronend after video, as the audio upload can block for a
long time and unnecessarily delay video frame presentation.

Fix this by accumulating all audio samples into a single output buffer
which is submitted at the end of each frame.

Reviewers
======
@sonninnos @jdgleaver


sonninnos:
Thanks! Works nicely here.

PR #529, opened 2022-05-24
Build fix: glibc 2.35+

repojohnray:
In file included from sources/src/audio.c:46:
/usr/include/bits/mathcalls.h:133:1: error: expected ';' before 'extern'
 __MATHCALL_VEC (log2,, (_Mdouble_ __x));

The following defines: log2 and log2l were interacting with the
commit: 7e1722fec84c65bf95f249f9ad8d15ab12d8c853 from glibc

### DIFF
--- a/sources/src/fpp_native.c
+++ b/sources/src/fpp_native.c
@@ -903,7 +903,7 @@ static void fp_log10(fpdata *a, fpdata *b)
 static void fp_log2(fpdata *a, fpdata *b)
 {
 	fp_normal_prec();
-	a->fp = log2l(b->fp);
+	a->fp = logbl(b->fp);
 	fp_reset_normal_prec();
 	fp_round(a);
 }
--- a/sources/src/include/sysdeps.h
+++ b/sources/src/include/sysdeps.h
@@ -746,11 +746,4 @@ typedef uint8_t uint8;
 #define _daylight 0
 #endif
 
-#ifndef log2
-#define log2 logb
-#endif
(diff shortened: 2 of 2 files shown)


sonninnos:
Those defines were added because some platforms (can't remember which) did not build without them. My current setup (Windows and Linux) builds fine though without them. And the pipeline is working, so why is this needed again?

And also I refuse changes to the codebase without `__LIBRETRO__` guards, so maybe there is a better way to handle this.

asciibeats:
Got the same error and this fixed it for me. Compiling for aarch64. Thanks @repojohnray!

LibretroAdmin:
Hi @repojohnray , I think @sonninnos is still waiting for a response. He had some questions as expressed above.

repojohnray:
Hi @LibretroAdmin, I've no idea why these defines were implemented; I confirm these defines are not needed with Linux.

octartis:
Fixed my build on Arch while building package `libretro-uae-git`

veganvelociraptor:
This should be merged as soon as possible. I've been unable to build for months without this pr.

sonninnos:
It won't be until the changes I requested are done, and until someone is willing to fix all the builds that will fail after merging.

PanAeon:
This also blocks build on `NixOS`

sonninnos:
I updated my Linux VM and finally got this failure too, and found another fix without editing `fpp_native.c`.

Please confirm.

PR #564, opened 2022-11-13
Makefile: fix linking for armv

SupervisedThinking:
- `-lm` is necessary to sucessfully link puae & `-lpthread -ldl` where needed for glibc < 2.34 too


sonninnos:
Thanks! Merging.

PR #618, opened 2023-07-18
Update 5.0.0

sonninnos:
Fresh WinUAE codebase.


zzedmore:
Just seen this. Thanks for the update to v5.0.0. i tested it with the performance hungry 'Jim Power' and it seems to be about 6-8 fps slower than the previous v4. This was on my oldish Android phone though. I think Toni Willen said he removed some hacks so maybe that's why the requirements have gone up slightly ? We still have the 2021 version so all is still good. Thanks.

sonninnos:
Thanks. More accurate tends to be more slow.

zzedmore:
Apart from accuracy, was there anything else missing from the 2021 version ?

sonninnos:
It has old CPU and custom chip stuff, but sound and everything else regarding core options should be on par. Except of course it only has one Cycle-exact mode, which is full. And it does not have lightgun or CD32 FMV support.

PR #647, opened 2024-02-19
Increase analog stick mouse precision to sub-pixel

rsn8887:
The analog mouse currently has a slowest speed given by the fact that the process_analog_mouse function can return only a minimum value of 1. The code automatically turns the analog mouse into a digital mouse if the motion would be smaller than an integer value of 1. This is still pretty fast motion and makes it almost impossible to precisely position the pointer over a single pixel using the analog stick. It makes playing a game like Dreamweb with the analog stick extremely frustrating, even when holding down the slow mouse hotkey.

This following part in the existing code turns a very slow, analog, stick mouse movement (which would result in an increment of less than 1) into a faster, digital, movement:

```
   if (!mouse_axis && abs(analog_axis) > analog_deadzone)
      mouse_axis = (analog_axis > 0) ? 1 : -1;
```

The solution is to not use the above code, but instead allow sub pixel accuracy internally, so we keep a more fine grained record of mouse motion between frames. This PR is attempting to solve this. I am unable to test at the moment, but I can test it later this week.

The same method is used in ScummVM to allow very very slow and precise pointer motion.

### DIFF
--- a/libretro/libretro-mapper.c
+++ b/libretro/libretro-mapper.c
@@ -899,14 +899,21 @@ static int adjust_analog_deadzone(int analog_axis, int analog_deadzone)
    return analog_adjusted;
 }
 
-static int process_analogmouse(int analog_axis, int analog_deadzone, int mouse_multiplier)
+static int process_analogmouse(int analog_axis, int analog_deadzone, float mouse_multiplier, float *sub_pixel_remainder)
 {
-   int analog_adjusted = adjust_analog_deadzone(analog_axis, analog_deadzone);
-   int mouse_axis      = 0;
+   int mouse_axis = 0;
 
-   mouse_axis = analog_adjusted * 10 * opt_analogmouse_speed / (32768.0f / mouse_multiplier);
-   if (!mouse_axis && abs(analog_axis) > analog_deadzone)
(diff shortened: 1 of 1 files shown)


rsn8887:
Note: This is untested currently.

rsn8887:
Just for the record, sonninnos tested this a few minutes ago and reported "twitchiness" in normal, regular speed, mouse motion. I made a small change to reset the hires coordinate variables whenever the axis value is zero. I will test later today to see if I can make any other improvements.

rsn8887:
I found another improvement and force-pushed it.

rsn8887:
Ok the trick is to reset the hires variables whenever no analogue input for the axis occurs. I forgot this in the original version.

rsn8887:
All problems fixed and tested on MacOS. Should be good to go.

sonninnos:
Feels ok, and as a minor nitpick those `process_analogmouse()` if-else hassles could be removed and just do it all in `process_analogmouse()`, but not that important.

sonninnos (approved):

rsn8887:
> Feels ok, and as a minor nitpick those `process_analogmouse()` if-else hassles could be removed and just do it all in `process_analogmouse()`, but not that important.

Good suggestion, thanks. I made those changes.

PR #655, opened 2024-05-04
Analog stick mouse refactor + cleanup

### LINKED ISSUE #654: [Feature Request] Diffenent mouse speed for left/right analog-sticks
I'm new to libretro/RetroArch so forgive me if i bother you. Also i'm not an english speaking native and hope you understand what i want! :)

I recently purchasted an Retroid Pocket 4 Pro and would like to emulate my Amiga-childhood. 
Trying to play Zak McKracken feel's a little bit strange, since there is no real mouse. 
Using the Analog-Sticks is also complicated since the mouse speed is to fast or to slow in my opinion. 
I know there is the posibitity to config a "mouse fast/slower" button. But i think there are two analog-sticks doing the same on my device. 
So why do not config one analog stick fast, the other one slow and add both values to control the mouse.

I'm a developer by myself, but only working with VS2022 and not really familiar with creating pull requests. 
But i can read your well developed code and would suggest changes in two files. I will ad them as attachment. Since i'm also to stupid to genate patchfiles i will also add the originals ("*.Gid.c - Extension")

Maybee the developers can include my code to the build. 
Or as an option point me to cross-compile-beginner-guide, so i can compile it myself for my Retroid.

[Amiga.zip](https://github.com/libretro/libretro-uae/files/15062363/Amiga.zip)

sonninnos:
- Add separate speed settings for both analog sticks
  Closes #654
- Allow both sticks to work together for more precision
- Refactor analog stick deadzone range adjustment
- Finetuned analog stick mouse sub pixel calculation
- Mouse slow toggle speed multiplier adjustment
- Dpad mouse fast toggle maximum speed correction
- Variable name clarifications and cleanups


rsn8887:
Why are you clamping the sub pixel remainder value? This shouldn’t be necessary because you are limiting minimum mouse speed now. There’s already a deadzone.

```
if (fabs(delta) < 0.1f)
         *sub_pixel_remainder = 0;
      else
         *sub_pixel_remainder += delta - mouse_axis;

      if (fabs(*sub_pixel_remainder) > 1.0f)
         *sub_pixel_remainder = 0;
```
It should work better without this (more fine grained control). Or if there’s some problem without this, there’s a bug somewhere else I’d say, because the above code is a hack.

sonninnos:
Before it clamped even bigger values. Please try it first.

rsn8887:
Ok I can try it later on Steamdeck, RPi and MacOS and report.

PR #688, opened 2025-01-22
Android workaround for capsimg

kadireksi:


sonninnos:
Please fix the messy whitespaces, and don't add unnecessary braces, and readme must be updated too..

kadireksi:
hope this is ok now )'

sonninnos:
Close enough, thanks, I'll clean it up later since have to update the old 2021 too.

PR #716, opened 2025-11-02
Fix loading of Storage Access Framework files from the libretro VFS interface

### LINKED ISSUE #715: libretro vfs support
Hey @sonninnos , not sure this is my screwup, but I've been experimenting with the new RetroArch support for SAF and RSAF to keep my games out of local storage, since I already had a read only local network with webdav for other programs, by using playlists pointing to files in the webdav (interestingly this works with m3u in the beetle psx core... even if they are using absolute paths, not relative with portable playlists. Not sure if it's the portable playlists carrying water there, but I changed all my m3u to use relative paths anyway, since I did a cmd line tool to create m3u anyway so it was easy).

As the coder implementing the feature said, some cores can boot remote games some don't and the difference is libretro VFS support in the core. I didn't expect puae to be one that didn't since you're so involved with libretro and I see commits in the first page referencing the libretro vfs, but I just tried to boot various retroplay whdloads and it didn't, just stay stuck at the kickstart 1.3 screen, while copied to local storage works. 

Just letting you know if you're not aware yet, or if it's something specific to whdload.

white-axe:
Closes #715.

### DIFF
--- a/sources/src/cfgfile.c
+++ b/sources/src/cfgfile.c
@@ -705,7 +705,14 @@ static TCHAR *cfgfile_get_multipath2 (struct multipath *mp, const TCHAR *path, c
 	for (int i = 0; i < MAX_PATHS; i++) {
 #ifdef __LIBRETRO__
 		/* Hacky fix for relative LHA paths, but HDFs work without.. */
-		if (!path_is_absolute(file) && path_is_valid(file))
+		bool has_scheme = false;
+		for (TCHAR *ptr = file; *ptr != 0 && *ptr != '/' && *ptr != '\\'; ++ptr)
+			if (*ptr == ':')
+			{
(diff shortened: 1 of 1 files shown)


sonninnos:
I merged it also in the 2.6.1 branch, and also noticed that it added a warning:
```
sources/src/cfgfile.c: In function 'cfgfile_get_multipath2':
sources/src/cfgfile.c:398:35: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  398 |                 for (TCHAR *ptr = file; *ptr != 0 && *ptr != '/' && *ptr != '\\'; ++ptr)
      |                                   ^~~~
```

white-axe:
Oops. The ptr variable can be const.

PR #731, opened 2026-05-12
fix: Correct fseek error check in my_fsize

garbear:
## Description

This change treats fseek failure as `< 0` in file size check.

The libretro VFS API requires fskeek to return "the new position": https://github.com/libretro/libretro-common/blob/7040de62e/include/libretro.h#L2853

## Motivation and context

`my_fsize()` was treating non-zero seek results as failure, but the wrapped seek API may return the new position on success.

To fix this, we check for negative return values instead to avoid false "fseek failed" errors and restore WHDLoad startup script execution.

## How has this been tested?

To reproduce:

> The file is visible to the Amiga but something goes wrong when trying to run it. If I type type "cd c" then "dir" in the Amiga shell, the shell returns "dir: file is not executable" and this line is generated in the log:

> 2026-04-12 16:22:31.255 T:1679796    info <general>: AddOnLog: game.libretro.uae: my_fsize: fseek on file '/Users/user/Library/Application Support/Kodi/userdata/addon_data/game.libretro.uae/save/WHDLoad/C/Dir' failed

Testing in progress here: https://github.com/kodi-game/game.libretro.uae/issues/3

### DIFF
--- a/retrodep/fsdb_host.c
+++ b/retrodep/fsdb_host.c
@@ -428,7 +428,7 @@ uae_s64 my_fsize(struct my_openfile_s* mos) {
 #else
 	size_t size = 0;
 	int current = ftell(mos->fp);
-	if (fseek(mos->fp, 0, SEEK_END)) {
+	if (fseek(mos->fp, 0, SEEK_END) < 0) {
 		write_log("my_fsize: fseek on file '%s' failed\n", mos->path);
 		return -1;
 	}


sonninnos:
Unfortunately the whole confusion is because the documentation and API is wrong about what seek is supposed to be. It was and is a replacement for `fseek` instead `lseek`, because that is what happens in `file_stream_transforms.h` so of course VFS variants ought to follow the same behavior, but somewhere along the way the current state of `retro_vfs_file_seek_internal()` is undecided and does oddly both, and from there it probably ended up in the documentation as if that is the desired state..

The part you changed however is when libretro VFS is disabled with `NO_LIBRETRO_VFS=1` so VFS stuff can't really play a role, and I also tried to repro the error, but could not in RetroArch with or without VFS.

garbear:
OK, will report back with more testing.

garbear:
So, in the issue I linked the user reported that this change fixed the problem. I haven't scrutinized `NO_LIBRETRO_VFS=1`, but it's the Kodi VFS that returns position, which we then pass back to the libretro API because it specifies returning the same position. What should we do about the change here? Does it need more scrutiny? It might be unfortunate that the libretro API chose a non-standard return value, tho Kodi also conforms to this, so returning a position must not be too uncommon in the ecosystem.

sonninnos:
Like I said it simply does not make any sense that this particular change would fix any replaced VFS stuff since it changes code that does not use VFS replacements but regular `fseek`, since it is outside of `#ifdef USE_LIBRETRO_VFS`. `fseek` does what the manual says:

> Upon successful
       completion, fgetpos(), fseek(), fsetpos() return 0, and ftell()
       returns the current offset.  Otherwise, -1 is returned and [errno](https://www.man7.org/linux/man-pages/man3/errno.3.html)
       is set to indicate the error.

Unless of course your frontend replaces those functions even when VFS is disabled in the core, which sounds very much wrong. The if currently fails when the return value is non-zero, so changing it to not fail on positive values can't be right.. Probably won't do harm here, but still can't be right. Especially since it works here either way, as in I can't replicate the problem.

It is `lseek` behavior:
> Upon successful completion, lseek() returns the resulting offset
       location as measured in bytes from the beginning of the file.  On
       error, the value (off_t) -1 is returned and [errno](https://www.man7.org/linux/man-pages/man3/errno.3.html) is set to
       indicate the error.


Edit: I mean sure just recently the code was the way that the change would make sense:
https://github.com/libretro/libretro-uae/commit/bd2ef50e22f5ded91cae347e98352f5bd2e7e6c2

And after that
https://github.com/libretro/libretro-uae/commit/e90166872acd0df1d9908c986139848e93c8f188

Meaning that this fix should not be needed anymore.

garbear:
You're right, this change shouldn't affect code outside `USE_LIBRETRO_VFS`. It's possible that https://github.com/libretro/libretro-uae/commit/e90166872acd0df1d9908c986139848e93c8f188 fixed the problem. Either way this fix shouldn't be needed.

Issue #729, opened 2026-04-30
un-named symbol


paigeadelethompson:
```
INFO] [Content] Content loading skipped. Implementation will load it on its own.
[INFO] [Environ] SET_PIXEL_FORMAT: XRGB8888.
[libretro INFO] Unzip: /home/sq/.config/retroarch/saves/PUAE/TEMP/Zak1.adf
[libretro INFO] Unzip: /home/sq/.config/retroarch/saves/PUAE/TEMP/Zak2.adf
[libretro INFO] ->M3U: /home/sq/.config/retroarch/saves/PUAE/TEMP/Zak McKracken.m3u
[libretro INFO] No model specified in: "/home/sq/.config/retroarch/saves/PUAE/TEMP/Zak McKracken.m3u".
[libretro INFO] Model: "A500".
[libretro INFO] Kickstart: "kick34005.A500".
[libretro INFO] M3U parsed, 2 file(s) found
[libretro INFO] Disk (1) inserted in drive DF0: "/home/sq/.config/retroarch/saves/PUAE/TEMP/Zak1.adf".
[libretro INFO] Disk (2) inserted in drive DF1: "/home/sq/.config/retroarch/saves/PUAE/TEMP/Zak2.adf".
[libretro INFO] --------------------------------------------------------------------------------
[libretro INFO] PUAE 5.3.1 e146c41b, Apr 30 2026 08:53:35
[libretro INFO] --------------------------------------------------------------------------------
Process 30398 stopped
* thread #1, name = 'retroarch', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
    frame #0: 0x000000084265fc44 libc.so.7`_sem_destroy [inlined] sem_check_validity(sem=0x0000000000000000) at cancelpoints_sem_new.c:116:18
(lldb) bt
* thread #1, name = 'retroarch', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
  * frame #0: 0x000000084265fc44 libc.so.7`_sem_destroy [inlined] sem_check_validity(sem=0x0000000000000000) at cancelpoints_sem_new.c:116:18
    frame #1: 0x000000084265fc44 libc.so.7`_sem_destroy(sem=0x0000000000000000) at cancelpoints_sem_new.c:314:6
    frame #2: 0x00000009171eb57a puae_libretro.so`___lldb_unnamed_symbol37504 + 122
[...]

PR #735, opened 2026-05-17
check if reference is null, and check if struct member is null before…

paigeadelethompson:
… calling sem_destroy

This seems to be at least a realistic enough fix for https://github.com/libretro/libretro-uae/issues/729#issuecomment-4469270823

at least this patch gets to a working game without removing filesys_cleanup or modifying anything in sysconfig.

### DIFF
--- a/retrodep/threaddep/thread.h
+++ b/retrodep/threaddep/thread.h
@@ -129,7 +129,9 @@ void uae_sem_init (uae_sem_t *sem, int pshared, unsigned int value);
 
 STATIC_INLINE void uae_sem_destroy (uae_sem_t *sem)
 {
-    sem_destroy (sem->sem);
+    if(sem && sem->sem) {
+        sem_destroy (sem->sem);
+    }
 }
 
 STATIC_INLINE void uae_sem_post (uae_sem_t *sem)


sonninnos:
Interesting. Why not add similar checks for the rest of the functions while at it. Except then again if there is no need then why bother.

paigeadelethompson:
give me a sec I'll see what I can do :)

paigeadelethompson:
Seem  I've made a mistake, 
```
commit d998eee5dbf653d3cdeeeedecccaa3cfea859701 (HEAD -> master)
Author: sonninnos <tony@pihaus.org>
Date:   Tue May 12 05:17:49 2026 +0300

    Threaddep cleanup
```

whatever was changed in that commit probably also fixed the issue better than I did; I noticed there was some issues when I was trying to create a pull the first time around and I saw this in there so my bad-- anyway I just rebuilt it from that commit `d998eee5dbf653d3cdeeeedecccaa3cfea859701` and it seems to work fine too so *shrug*

paigeadelethompson:
Well OK I don't know that there's anything that needs to be done here necessarily ... but like you said lets assume it made sense and this didn't get fixed in a commit that I just didn't have .. 

```
if(sem->sem)  {
```
this sorta makes the assumption that sem is not NULL itself, there's also 

```
if(! sem->sem) { 
  return -1;
}
```

few thoughts: 
- `if(!sem || !sem->sem)` or `if(sem && sem->sem)` might be a good idea with exception to `return sem->sem == 0 ? -1 : sem_trywait(sem->sem);` it's probably fine to just check `if(sem)` 
- not everything returns a lot of these are `void` so maybe it would be worthwhile to do: 

```
if(sem && sem->sem) { 
    // ... 
} else { 
    bug(_T("uae_sem_fn: sem or sem->sem is null"); 
}
```
and that might at least give some context for whats going on but I don't know if `bug` from `debug.h` would be the right thing to use or not; it's an alias for `write_log` if I'm not mistaken. If you want I could do something like that and re-send a new pull request but FWIW I don't think this one is necessary anymore.

sonninnos:
> whatever was changed in that commit probably also fixed the issue better than I did; I noticed there was some issues when I was trying to create a pull the first time around and I saw this in there so my bad-- anyway I just rebuilt it from that commit `d998eee5dbf653d3cdeeeedecccaa3cfea859701` and it seems to work fine too so _shrug_

Good, and odd, since the commit did not even have the exact same change, but I can still add it.

Edit: Ok, pushed some more cleanups, so please test if I missed something.

paigeadelethompson:
> > whatever was changed in that commit probably also fixed the issue better than I did; I noticed there was some issues when I was trying to create a pull the first time around and I saw this in there so my bad-- anyway I just rebuilt it from that commit `d998eee5dbf653d3cdeeeedecccaa3cfea859701` and it seems to work fine too so _shrug_
> 
> Good, and odd, since the commit did not even have the exact same change, but I can still add it.
> 
> Edit: Ok, pushed some more cleanups, so please test if I missed something.

I’ll take a look soon sorry I missed this

sonninnos:
So is the problem gone and we can close this?

paigeadelethompson:
Yeah you can close sorry

